PR #2, opened 2014-12-17
Routing table

mmmries:
Looking for feedback about the direction I'm moving here. I want to get rid of the class level variables that track all of the rabbitmq connection, channels, configuration and queues. I want to replace these class-level variables with instance variables that live inside a separate object so that a set of subscriptions can be started and stopped.

This also allows me to create custom routes.

My plan for these changes would include:
  * the route object should contain a middleware stack, so we can customize the stack per-route
  * handle message acknowledgements (at_most_once! and at_least_once!) via middleware items, instead of calling env.acknowledge and env.reject in the existing middleware objects
  * keep the existing code paths working while I work towards a fully functioning SubscriptionSet over the course of multiple pull requests
  * deprecate the <code>auto_pop!</code> pattern. Now that we have at_most_once! working with a prefetch, we no longer need to support the old "pounce" mode
  * once SubscriptionSet is ready for production use I will change the <code>ActionSubscriber.start_subscribers</code> call so that it builds and starts a subscription set instead of using the existing class-level <code>auto_subscribe!</code> methods
  * remove a whole lot of dead code once we no longer need the old code paths

Any and all feedback is welcome here
/cc @quixoten @abrandoned @liveh2o @brianstien


mmmries:
Welcome back from the holidays :santa: :bell: :santa:

Here is a giant PR that needs some review!

mmmries:
not even I have been looking at this PR since it became so monstrously huge :japanese_ogre:

I'm going to close it and see if I can get some these ideas built into the current gem in a more iterative fashion

PR #3, opened 2015-01-21
better integration tests

mmmries:
I want to support having multiple integration tests so that I can get good coverage of all our major features like supporting custom content types, at_least_once!, at_most_once! etc.
Here I have attempted to simplify the process of how an integration test works by keeping the details in spec/spec_helper.rb
Currently the process for cleaning up at the end of a test run is pretty gross, but I just wanted a proof of concept with two integration tests before I add the rest of the test coverage.
I also removed simplecov since I never use it

/cc @quixoten @brianstien @abrandoned @liveh2o

### DIFF
--- a/action_subscriber.gemspec
+++ b/action_subscriber.gemspec
@@ -34,5 +34,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pry-nav"
   spec.add_development_dependency "rspec", "~> 3.0"
   spec.add_development_dependency "rake"
-  spec.add_development_dependency "simplecov"
 end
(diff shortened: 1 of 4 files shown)


mmmries:
BTW when we are only changing tests, do we release a point version of the gem? Seems like maybe we should wait until we have actually changed functionality, but I would like to keep the flow of feedback from running production systems to keep confidence high that these changes don't introduce any regressions.

In other words I don't want to wait a long time before releasing a gem update because I will be more nervous about the new gem version having a lot of differences from the last release.

liveh2o:
Spec changes won't require a release because they don't impact API or behavior. If a spec change leads to or accompanies an API or behavioral change, then a release should occur.

liveh2o:
:+1:

PR #5, opened 2015-01-22
Decoder coverage

mmmries:
Adds an integration test for the default JSON decoder as well as adding a custom decoder.

I also moved the `require "spec_helper"` call to .rspec which is the standard thing rspec does today if you run `rspec --init`

### DIFF
--- a/spec/integration/basic_subscriber_spec.rb
+++ b/spec/integration/basic_subscriber_spec.rb
@@ -27,6 +27,7 @@ describe "A Basic Subscriber", :integration => true do
       exchange = channel.topic("events")
       exchange.publish("Ohai Booked", :routing_key => "greg.basic_push.booked")
       exchange.publish("Ohai Cancelled", :routing_key => "basic.cancelled")
+      sleep 0.1
 
       ::ActionSubscriber.auto_pop!
 
--- a/spec/integration/basic_subscriber_spec.rb
+++ b/spec/integration/basic_subscriber_spec.rb
@@ -1,5 +1,3 @@
-require 'spec_helper'
-
 class BasicPushSubscriber < ActionSubscriber::Base
   publisher :greg
 
--- a/spec/integration/decoding_payloads_spec.rb
+++ b/spec/integration/decoding_payloads_spec.rb
@@ -1,0 +1,48 @@
+class TwitterSubscriber < ActionSubscriber::Base
+  def tweet
+    $messages << {
+      :decoded => payload,
+      :raw => raw_payload,
+    }
+  end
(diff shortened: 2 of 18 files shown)


spec/integration/basic_subscriber_spec.rb
```diff
@@ -27,6 +25,7 @@ def cancelled
       exchange = channel.topic("events")
       exchange.publish("Ohai Booked", :routing_key => "greg.basic_push.booked")
       exchange.publish("Ohai Cancelled", :routing_key => "basic.cancelled")
+      sleep 0.1
```
mmmries: I never ran into this problem locally with jruby-1.7.18 or jruby-9k, but on travis they were failing alot in a way where they didn't get the messages back from rabbitmq in this test. I think the issue was that it tried to pop right after the publish, but the publish is async so the messages weren't ready in the queues yet (I think)

So I added this sleep and now the travis tests seem to pass just fine...

spec/integration/decoding_payloads_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    exchange.publish(json_string, :routing_key => "twitter.tweet", :content_type => "application/json")
+    sleep 0.1
+
+    expect($messages).to eq Set.new([{
+      :decoded => JSON.parse(json_string),
+      :raw => json_string,
+    }])
+  end
+
+  it "can be given custom decoders" do
```
liveh2o: I'd prefer to tie the adding of a custom decoder directly to this spec. When I looked through it initially (on my phone), it wasn't clear where the custom decoder was being defined, which also made it more difficult to decipher how the expected decoded value was being determined. Perhaps we can even unload/reset it after the specs run to ensure we are getting an isolated test case?

spec/integration/decoding_payloads_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    expect($messages).to eq Set.new([{
+      :decoded => JSON.parse(json_string),
+      :raw => json_string,
+    }])
+  end
+
+  context "Custom Decoder" do
+    let(:content_type) { "foo/type" }
+
+    before { ::ActionSubscriber.config.add_decoder(content_type => lambda{ |payload| :foo }) }
```
mmmries: is this clearer? It sort of provides isolation, but hopefully once I start refactoring the internals we can get this isolation without having to clobber global state

mmmries:
I squashed the commit to clarify the decoder spec, mostly because I'm trying to get into the habit of using `git rebase -i`

PR #6, opened 2015-01-23
Connection recovery test

mmmries:
Adds an integration test for automatically recovering from dropped connections.

Note: this test is significantly slower than the others so I tagged it with `:integration => true, :slow => true`. The test will still be executed on the CI server but you can skip it during normal development by running `rspec --tag ~slow` or just run the unit tests with `rspec spec/lib`

I also updated the README to reflect some of the recent changes about how to configure ActionSubscriber

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,19 +70,21 @@ ActionSubscriber needs to know how to connect to your rabbit server to start get
 In an initializer, you can set the host and the port like this :
 
     ActionSubscriber::Configuration.configure do |config|
-      config.host = "my rabbit host"
+      config.hosts = ["rabbit1", "rabbit2", "rabbit3"]
       config.port = 5672
     end
 
 Other configuration options include :
 
+* config.add_decoder - add a custom decoder for a custom content type
 * config.allow_low_priority_methods - subscribe to queues for methods suffixed with "_low"
 * config.default_exchange - set the default exchange that your queues will use, using the default RabbitMQ exchange is not recommended
-* config.hosts - an array of hostnames in your cluster
-* config.times_to_pop - when using RabbitMQ's pull API, the number of messages we will grab each time we pool the broker
-* config.threadpool_size - set the number of threads availiable to action_subscriber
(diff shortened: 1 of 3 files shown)


liveh2o:
Does Travis run the specs with Rake, or does it use RSpec directly? If it uses RSpec, then I suggest changing the rake file so that only specs under lib/ are run and adding a task that runs the entire suite, including integration/. Thoughts?

mmmries:
Travis just runes `rake` by default, but we can override it to run whatever command we want.

Just to confirm what you want to have happen:
* when a user runs `rake` or `rspec` it should skip integration tests
* Travis should run all tests include integration tests

Is that the behavior you are looking for?

liveh2o:
I guess, but that's only because you said the spec as slow. It seems like we should optimize for the most common use case.

mmmries:
Here are the timings:
* `spec/lib` Finished in 0.12541 seconds (files took 0.45364 seconds to load)
* `spec/integration` Finished in 6.33 seconds (files took 0.43213 seconds to load)
* `spec/integration/automatic_reconnect_spec.rb` Finished in 5.33 seconds (files took 0.46 seconds to load)
* `all` Finished in 6.69 seconds (files took 0.8906 seconds to load)

To me that is still fast enough that I would prefer to have everyone run all the specs on a regular basis. The reconnect test is by far the slowest test in the overall test suite, but I think 6.7sec is not a heavy cost to impose. If the integration suite continues to get slower (due the fact of having to interact with rabbitmq) then I can switch it to default to just the unit tests and make travis run the full quite, but for now I vote we keep it simple and just make the full suite be the default for everyone.

mmmries:
I modified the reconnect test so that instead of just always sleeping 5 seconds it checks the connection status 10 times per second and sleeps up to 10 seconds. The test now averages about 2.5sec instead of 5.5esc to run. This makes it even less of a burden to run this test all the time.

liveh2o:
~5 seconds isn't bad. Let's roll as is. :shipit:

PR #7, opened 2015-01-27
get rid of the custom runner

mmmries:
`ActionSubscriber::Middleware::Runner` was just verifying that router was on the top of the stack, but we can just use insert_before to keep router on the top of the stack.

Our README does not document anything about modifying the middleware stack, so I think `ActionSubscriber.config.middleware` is currently considered a private API. In the near future I want to remove the idea of a single global middleware and allow each subscriber (or each subscription) to have its own middleware, but for now I just want to get rid of this non-standard use of a runner.

/cc @localshred @quixoten @abrandoned @brianstien


mmmries:
BTW despacho is already using `insert_before` so we shouldn't have any issues there

liveh2o:
I disagree with removing the "non-standard" runner. The fact that the middleware gem doesn't provide the ability to set a terminating middleware is a bug (IMO). In Rails, for example, you can't add anything after the routing middleware. You don't need to be aware of this to use your own middleware, and you can safely use the `use` method to push your custom middleware on to the stack without interfering with the stack. By not providing this alternate runner here, it's not clear at all that you would need to use `insert_before` when pushing something onto the stack.

The fact that README doesn't document this is not indicative of it being considered private, but rather laziness/forgetfulness on the part of the implementer (me). In fact, I added it so that we could hook into this process without interfering with what was going on in Action Subscriber itself. That's actually why I chose to use the middleware pattern in the first place.

mmmries:
I've been thinking about forking the middleware gem anyway because I really want the ability to create copies of a middleware stack (ie copy the global ActionSubscriber.config.middleware stack and then customize a copy for each queue), which it also doesn't support and the maintainer doesn't seem interested in issues or pull requests.

So maybe we just fork it and scratch our own itch?

liveh2o:
Yeah, I'd be down. We could call it Malcolm.

mmmries:
:heart: :heart: :heart: :heart_eyes: I love that show and that name. Commencing forking...

PR #8, opened 2015-01-30
support nested around filters

mmmries:
I'm adding the idea of `around_filters` for a subscriber. This is a stepping stone for me to move the implementation of `at_least_once!` and `at_most_once!` modes into filters rather than having the implementation being spread throughout the codebase.

For now I'm not adding any other options to these filters (ie. `:only => [:this_action]` or `:except => [:that_action]`), because I don't need those features so I just made it as simple as I could.

### DIFF
--- a/lib/action_subscriber/dsl.rb
+++ b/lib/action_subscriber/dsl.rb
@@ -90,10 +90,10 @@ module ActionSubscriber
 
     def run_action_with_filters(env, action)
       subscriber_instance = self.new(env)
-      final_block = ->{ subscriber_instance.public_send(action) }
+      final_block = Proc.new { subscriber_instance.public_send(action) }
 
       first_proc = around_filters.reverse.reduce(final_block) do |block, filter|
-        ->{ subscriber_instance.send(filter, &block) }
+        Proc.new { subscriber_instance.send(filter, &block) }
       end
       first_proc.call
     end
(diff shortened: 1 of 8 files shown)


lib/action_subscriber/dsl.rb
```diff
@@ -79,5 +87,15 @@ def routing_key_for(method, routing_key_name)
     def routing_key_names
       @_routing_key_names ||= {}
     end
+
+    def run_action_with_filters(env, action)
+      subscriber_instance = self.new(env)
+      final_block = ->{ subscriber_instance.public_send(action) }
```
liveh2o: Stabby procs?

archer-puts-lana-in-the-danger-zone-gif

mmmries: They need to be procs so that I can have a closure around the subscriber instance. The scope of their existence is very limited so I'm not super concerned about causing memory leaks.

Also this is the same pattern that [protobuf](https://github.com/localshred/protobuf/blob/c0257ec32e4da9b3979dfd5dc6d554ff0c6bffe8/lib/protobuf/rpc/service_filters.rb#L212) uses for its around filters

liveh2o: I'm fine with using procs. Just not the stabby proc syntax 😀

mmmries:
changed stabby procs to `Proc.new { ... }`. I had never noticed that part of our ruby style guide before. Can anyone point me to the reasoning of why we prefer `Proc.new { }` over `-> { }`?

liveh2o:
:shipit:

PR #9, opened 2015-01-30
at_least_once! and at_most_once! implemented as filters

mmmries:
this is a refactoring of our implementation of `at_least_once!` and `at_most_once!` modes using filters instead of using the middleware

### DIFF
--- a/lib/action_subscriber/base.rb
+++ b/lib/action_subscriber/base.rb
@@ -76,6 +76,19 @@ module ActionSubscriber
       env.acknowledge
     end
 
+    def _at_least_once_filter
+      yield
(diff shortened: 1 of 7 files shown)


mmmries:
I would like to merge this in and release it together with the `around_filter` changes. Can I get a quick peak?

localshred:
Is this the part where I give a no-look shipit? :)

mmmries:
@localshred I accept all major forms of shipits. But I prefer the kind where you checkout the diff first :eyes: :eyeglasses:

liveh2o:
:shipit:

PR #12, opened 2015-09-09
Brett/rollup

localshred:


localshred:
:+1:

mmmries:
:+1: to the idea of collapsing babou into this project!

mmmries:
Even better if we can get rid of the name Babou and set all the configuration options directly from the `action_subscriber` executable

Issue #14, opened 2015-09-10
Addtional Features: Action Publisher & Babou


brettallred:
What has the conversation been regarding rounding out this gem as an end to end solution rather than only part of the messaging solution.

This could include
* Publishable support for an ActiveRecord Model
* Native Subscriber processor (Babou) in the same way that Promiscuous has their subscriber built in rather than in a different repo.
* Update RabbitConnection to support Rx & Tx connections for separating out connections for consumers and publishers

I have a proof of concept working internally that does this as a single package for event messaging but was curious the direction you guys are taking on this?

mmmries:
> Native subscriber process..

This should definitely be built into this gem.

> Publishable support for an ActiveRecord model

I had planned on creating a `ActionPublisher` gem because there are some apps we have internally that need to publish, but don't need to subscribe to anything. But maybe it doesn't cost us much to just load the publishing code from this gem? The dependencies will be pretty much the same. So perhaps a second gem is not very useful at all.

> Update RabbitConnection

I don't think we need to separate the two connections. When we publish we create a new channel within the connection, publish from that channel and then shut the channel down. Does anyone know of a good reason to keep separate connections for publishing vs subscribing?

mmmries:
PS: Really happy someone outside of MX is interested in using this gem :heart: :heart_eyes: :boom: :zap: :tada:

brettallred:
RE: Separating consumers and publishers

Here a few articles from RabbitMQ and then a blog post from a cloud provider of RabbitMQ on how & why they do it.

http://www.rabbitmq.com/flow-control.html
http://www.rabbitmq.com/blog/2014/04/14/finding-bottlenecks-with-rabbitmq-3-3/
http://bigwig.io/docs/message_throughput/

mmmries:
Interesting. Looks like I need to instrument something on our rabbit cluster to watch for `flow` states and `consumer utilization`.

That sounds like a valid reason to me to have a separate publishing connection. Perhaps we should change `ActionSubscriber::RabbitConnection.connect!` to be something like `Action::Subscriber::RabbitConnection.connect_for_subscriptions!` and `Action::Subscriber::RabbitConnection.connect_for_publishing!`? Each of those methods could synchronize the lazy loading of a single connection?

brettallred:
> Native subscriber process

Are you guys okay with open sourcing babou and just using that as the internal subscriber process?  I am happy to put up a PR with that change integrated.

mmmries:
[...]

PR #17, opened 2015-09-11
Separate connections

mmmries:
Part of #14

Create a separate interface for handling a publishing connection from the subscriber connection. I still need to add the actual publishing code and an ActiveRecord mixin, but this starts us down the path towards having a separate connection for publishing and subscribing.

/cc @brettallred @localshred @liveh2o @abrandoned


brettallred:
Looks pretty straight forward to me :+1:

PR #18, opened 2015-09-12
Add Babou

brettallred:
This implementation currently works but there are a few questions below and things to do before merging.

Completed

* Add Babou
* Modify CLI to be executed by ~action_subscriber start rather than ~babou start
* Keeps original babou configuration.

Questions / To Do

* Ideas for naming.  Internally do we keep the name babou or rename it to something more generic?
* Merge mode, pop_interval, and  times_to_pop into the action subscriber configuration file so there are not two separate configurations?
* Write tests against babou.  I don't believe there have ever been tests written against it.

//cc @mmmries @liveh2o @abrandoned @brianstien


mmmries:
> Ideas for naming

In the long run I think we rename `babou` to something generic like `action_subscriber/cli`, but we can easily leave that for a future PR. It shouldn't change the external interface of the gem at all.

> Merge mode, pop_interval, times_to_pop

:+1: definitely merge the two configuration objects into a single object

> Write tests against babou

I have no strong desire to test the babou code. It is a pretty minimal set of functionality that just wraps around the interface that is already tested in the `spec/integration` tests. The only thing I think we should test is the code that drains the current worker threads before exiting the process, but that should probably be brought out of the CLI code and into the core of the library to make it more testable.

mmmries:
I'm also noticing right now that the behavior of draining the queues before exiting the process only works for pounce mode. We should investigate making that same behavior work in prowl mode to avoid losing messages during process shutdown.

mmmries:
This might also be a good time to rename `pounce` and `prowl` to something that we can actually remember. I still can't ever remember which mode is which without looking at the source code.

film42:
I just wanted to :+1: the renaming of `pounce` and `prowl`.

mmmries:
Instead of a single `start` endpoint with two different modes (`pounce` and `prowl`), what if we have two different CLI commands?

`action_subscriber poll` and `action_subscriber connect`? I don't love the name `connect`, but I think it would really helpful to make this into two separate thor actions so we can have `times_to_pop` and other configuration settings only be available for one action or the other.

film42:
@mmmries I was thinking of something along the lines of `push` or `subscribe`. Maybe those convey the wrong idea, but that's essentially what's happening under the hood. Perhaps we could event just call it `start`.

brettallred:
[WIP UPDATE]

* Configurations have been merged.  Unnecessary conifguration methods removed babou and babou CLI
* Prowl was renamed to subscribe and is the default option
* Pounce was renamed to pop and must still be set by the --mode  or by the yaml file. (open to further naming changes)
* Configuration from YML and CLI moved to ::ActionSubscriber::Configuration.  Later this will allow ActionSubscriber to be configured from a yaml file in addition to initializer (if wanted)




//cc @mmmries @film42 @abrandoned @brianstien

brettallred:
This finishes this PR.

bin/action_subscriber
```diff
@@ -0,0 +1,80 @@
...
+    class_option :app, :default => "./config/environment.rb"
+    class_option :mode
+    class_option :host
+    class_option :hosts
+    class_option :pop_interval, :type => :numeric, :desc => "how long to wait between asking for messages (in milliseconds)"
+    class_option :port, :type => :numeric
+    class_option :threadpool_size, :type => :numeric
+    class_option :times_to_pop, :type => :numeric, :desc => "how many messages to get from each queue each time we ask rabbit"
+
+    desc "start", "Start the babou subscription server"
```
mmmries: These description strings still use `babou` language. Let's change them for consistency.

bin/action_subscriber
```diff
@@ -0,0 +1,80 @@
...
+      require 'babou'
+
+      puts "Down Babou! ... down!"
+      puts "ROAAAAAARRRRRRRR!!!!!!"
+      wait_loops = 0
+      ::Babou.stop_server!
+
+      # Going to wait until the thread pool drains or we wait for 1000 seconds
+      # Only waiting for shut down in pounce mode
+      if ::Babou.pounce?
```
mmmries: `Babou` is no longer a top-level module, should be `ActionSubscriber::Babou`, also we should change this `pounce` to `pop`

bin/action_subscriber
```diff
@@ -0,0 +1,80 @@
...
+        end
+      end
+
+      exit 0
+    end
+  end
+end
+
+trap(:TTIN) {
+  require 'action_subscriber/babou'
```
mmmries: don't need to require babou here since we are using the `print_subscriptions` on the main `ActionSubscriber` module.

lib/action_subscriber/babou.rb
```diff
@@ -0,0 +1,103 @@
+module ActionSubscriber
+  module Babou
+    ##
+    # Class Methods
+    #
+    def self.configure_from_yaml_and_cli(cli_options = {})
+      env = ENV["RAILS_ENV"] || ENV["RACK_ENV"] || ENV["APP_ENV"] || "development"
+
+      yaml_config = {}
+      babou_absolute_config_path = ::File.expand_path(::File.join("config", "babou.yml"))
```
mmmries: Does anyone else think it would be worth searching for an `action_subscriber.yml` file and falling back to `babou.yml` if we don't find one?  I'm thinking that supporting `babou.yml` is nice for backwards compatibility, but we should be sticking with the ActionSubscriber name for new users.

film42: @mmmries Why not just rename this to `action_subscriber.yml` right now? I think it would be easy to change the config when we switch the CLI away from babou. Doesn't really make sense to support code that was never public.

mmmries: I only meant that it would be nice to have the backwards compatibility for
our own internal use so I could move updates up the pipeline even before we
get the new config files in place.

brettallred:
@mmmries - This fixes everything per your comments.

brettallred:
@mmmries - After merging 1.05 I am getting errors at

https://github.com/moneydesktop/action_subscriber/blob/master/lib/action_subscriber.rb#L68
https://github.com/moneydesktop/action_subscriber/blob/master/lib/action_subscriber.rb#L74

Looks like it got missed in the last update

mmmries:
This all looks good to me. Tomorrow morning I'll run a few of our apps against this version and use the new binstub just to be safe. As long as that goes well I'm a :+1: for merging and releasing.

mmmries:
Trying this out locally looks good. What do we think of releasing this as a `1.1.0-rc0` version? That way I can release it all the way up our chain in one or two apps before we give it the full go ahead and release it officially as `1.1.0`.

/cc @liveh2o @abrandoned @brettallred @localshred

film42:
:+1:

mmmries:
This has been released as `1.1.0.pre.rc0` (both MRI and jRuby version) https://rubygems.org/gems/action_subscriber/versions/1.1.0.pre.rc0-java

brettallred:
:smile:

lib/action_subscriber/configuration.rb
```diff
@@ -5,30 +5,64 @@ class Configuration
...
+    ##
+    # Class Methods
+    #
+    def self.configure_from_yaml_and_cli(cli_options = {})
+      env = ENV["RAILS_ENV"] || ENV["RACK_ENV"] || ENV["APP_ENV"] || "development"
+
+      yaml_config = {}
+      absolute_config_path = ::File.expand_path(::File.join("config", "action_subscriber.yml"))
+      if ::File.exists?(absolute_config_path)
+        yaml_config = ::YAML.load_file(babou_absolute_config_path, :safe => true)[env]
```
mmmries: `babou_absolute_config_path` should be `absolute_config_path` here

mmmries:
We've been running this code in production for several days now and have been running from the new binstub for several hours.

Everything so far looks really good. I don't see any reason not to merge this.

@liveh2o I think we should merge this branch and then remove the `ruby 2+` requirement in the gemspec and relase this as version `1.1`. That way we can go back to locking down the `amq-protocol` gem where necessary, but stay up to date with this gem. Thoughts?

Issue #22, opened 2015-09-17
Publishing


mmmries:
Part of #14 is to add publishing support.

Offline @liveh2o expressed a desire to either not include publishing in this gem or to rename the gem to something that isn't subscribing specific.

Here are some pros and cons I have been thinking about:
* The only direct dependencies required by both the publishing and subscribing code are `bunny` or `march_hare` depending on the platform. Releasing both an MRI and jRuby version of this gem is an annoying maintenance cost and I would rather not pay that cost twice.
* The only connection between publishing and subscribing is the naming convention so putting them in the same gem doesn't buy us much in terms of keeping the two gems in sync.
* `ActionSubscriber::Publisher` is kind of a weird name

The tradeoff that makes the most sense to me would be to have just a single `ActionSubscriber::Publisher.publish` call supported in this library. It will deal with the difference between `bunny` and `march_hare`, but won't include any support for detecting the content type or encoding the message.

That way most of the publishing logic that we currently have in our internal `Buttress` gem would stay there, it would only the platform-specific differences that are supported by this gem since that is a big part of its overall reason for existing.

Thoughts?
/cc @ah @abrandoned @brettallred @localshred @quixoten

mmmries:
This was resolved by #23

PR #23, opened 2015-09-24
Publishing

mmmries:
This is an implementation of my idea from #22 

The idea here is that we support a very basic form of publishing. ActionSubscriber handles the connection and a few compatibility issues between bunny and march_hare, but we aren't tackling encoding, content-types, etc.

For MX this would allow us to keep Buttress::Syncable and our protobuf encoding stuff intact, but remove all of our `if jRuby` conditionals. Buttress would no longer need to release jruby versions of that gem.

We could then create an `ActionPublisher` or perhaps `ActiveSyncable` gem that provides similar functionality in a more generic way so we can open source it.

/cc @brettallred @liveh2o @abrandoned @localshred

### DIFF
--- a/spec/integration/basic_subscriber_spec.rb
+++ b/spec/integration/basic_subscriber_spec.rb
@@ -21,10 +21,8 @@ describe "A Basic Subscriber", :integration => true do
 
   context "ActionSubscriber.auto_pop!" do
     it "routes messages to the right place" do
-      channel = connection.create_channel
-      exchange = channel.topic("events")
-      exchange.publish("Ohai Booked", :routing_key => "greg.basic_push.booked")
-      exchange.publish("Ohai Cancelled", :routing_key => "basic.cancelled")
+      ::ActionSubscriber::Publisher.publish("greg.basic_push.booked", "Ohai Booked", "events")
+      ::ActionSubscriber::Publisher.publish("basic.cancelled", "Ohai Cancelled", "events")
 
       verify_expectation_within(2.0) do
         ::ActionSubscriber.auto_pop!
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -93,3 +93,7 @@ module ActionSubscriber
 end
 
 require "action_subscriber/railtie" if defined?(Rails)
+
+at_exit do
+  ::ActionSubscriber::RabbitConnection.publisher_disconnect!
+end
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -20,6 +20,7 @@ require "action_subscriber/subscribable"
 require "action_subscriber/bunny/subscriber"
 require "action_subscriber/march_hare/subscriber"
 require "action_subscriber/babou"
+require "action_subscriber/publisher"
 require "action_subscriber/threadpool"
 require "action_subscriber/base"
 
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -11,11 +11,12 @@ module ActionSubscriber
                   :pop_interval,
                   :port,
                   :prefetch,
+                  :publisher_confirms,
                   :threadpool_size,
                   :timeout,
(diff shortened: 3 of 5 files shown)


Gemfile
```diff
@@ -2,3 +2,9 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in action_subscriber.gemspec
 gemspec
+
+if RUBY_ENGINE == "ruby" && RUBY_VERSION.split(".").first.to_i < 2
```
mmmries: @liveh2o I found out that the whole requirement for ruby 2.0+ only exists for MRI. Under jRuby there is no requirement on the `amq-protocol` gem, because `march_hare` includes its own implementation of the rabbitmq protocol.

We could definitely pull this out of this branch and put it into master with a patch release. That would resolve our 1.9 support issues.

spec/integration/basic_subscriber_spec.rb
```diff
@@ -21,10 +21,8 @@ def cancelled
 
   context "ActionSubscriber.auto_pop!" do
     it "routes messages to the right place" do
-      channel = connection.create_channel
-      exchange = channel.topic("events")
-      exchange.publish("Ohai Booked", :routing_key => "greg.basic_push.booked")
-      exchange.publish("Ohai Cancelled", :routing_key => "basic.cancelled")
+      ::ActionSubscriber::Publisher.publish("greg.basic_push.booked", "Ohai Booked", "events")
+      ::ActionSubscriber::Publisher.publish("basic.cancelled", "Ohai Cancelled", "events")
```
mmmries: We could switch all the other integration tests to publish this way, but I wanted to keep this PR really small for now and just have a single example of using it in an integration test.

Gemfile
film42: Can we bring this back into the gemspec, or do you want to leave this in the Gemfile?

film42:
This is looking really good. I'm all for :shipit: s, but you might want to wait for @liveh2o give the final (:+1:/:-1:).

Gemfile
mmmries: @film42 I don't think we can pull this back into the gemspec. The gemspec is evaluated when we are building the gem. So we can't have a version of the gemspec specific for MRI 1.9. But we should probably document that if you need to run MRI 1.9.X series they will have to add these specifications themselves

mmmries:
rebased this against master since I added back support for MRI 1.9 and jruby running under 1.9 mode back to the `1.1` series

liveh2o:
I'm definitely a fan of releasing this as a separate gem or renaming this gem to something that encompasses pub/sub. I would think that the binstub would still be called `action_subscriber`, but this way we aren't adding something action subscriber only to get publishing functionality.

film42:
We could always pull this into an `active_publisher` gem right now and include sync-ability into that.

mmmries:
I put together a quick `active_publisher` gem as a proof-of-concept (https://github.com/mxenabled/active_publisher).

Looking at the code side-by-side I still prefer managing the connection and basic publishing code in this project rather than duplicating a lot of the connection and configuration code between the two projects.

The ability to publish a message from a library called `ActionSubscriber` doesn't seem very weird to me. Anyone else want to weigh in on this?

localshred:
Having a publisher and subscriber in the same library seems perfectly normal, name notwithstanding. By colocating both modules you have the ability to share behavior (such as serializing/deserializing different content types). Setting up a separate gem for publishing seems overkill to me. ¯\\\_(ツ)\_/¯

mmmries:
@liveh2o and I chatted offline about this. Long term it might make sense for us to extract the `bunny`/`march_hare` logic into a library that can be shared between `action_subscriber` and `active_publisher`, but we don't have enough information right now to make that tradeoff so we are going to move forward with having basic publishing logic here, then we can have `active_publisher` move forward with a dependency on this gem and get some more information as we do that to decide whether it is worth splitting out.

film42:
cash_money

PR #27, opened 2015-10-01
add flag for allow-low-priority-methods

mmmries:
After deploying this into our production servers I realized that we needed an easy way to turn on/off the low priority methods/queues.

This just adds a command line option that gets passed into the configuration object to turn this feature on or off.

/cc @brettallred @localshred @liveh2o @abrandoned

### DIFF
--- a/action_subscriber.gemspec
+++ b/action_subscriber.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   end
   spec.add_dependency 'lifeguard'
   spec.add_dependency 'middleware'
+  spec.add_dependency 'thor'
 
   spec.add_development_dependency "activerecord", ">= 3.2"
   spec.add_development_dependency "bundler", ">= 1.6"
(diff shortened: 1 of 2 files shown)


action_subscriber.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   end
   spec.add_dependency 'lifeguard'
   spec.add_dependency 'middleware'
+  spec.add_dependency 'thor'
```
mmmries: I was testing the commandline outside of a rails app to make sure the commandline arg passes through correctly to the configuration object and realized that we don't have a declared dependency on thor.

We are clearly depending on it though so I added this explicit dependency

bin/action_subscriber
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env ruby
 
-require 'active_support'
-require 'active_support/core_ext'
 require 'thor'
+require 'action_subscriber'
```
mmmries: similar to the thor issue, this was working previously because rails had already loaded our gem, but I thought it was worth fixing so that the binstub just loads thor and the gem itself to make it less dependent on the behavior of what happens when we require `config/environment.rb`

film42:
:shipit:

PR #28, opened 2015-10-05
let the app load action_subscriber so we can check `if defined?(Rails)`

mmmries:
In the last PR (#27) I broke the way we load the `lib/action_subscriber/railtie.rb` file because we try to require it before we have required the `app` file (`config/enviornment.rb`).

This just reverts to the old behavior which lets the app decide what order to load things.

@ztoolson @ryanbjones @liveh2o @abrandoned


ztoolson:
:shipit:

PR #29, opened 2015-10-13
we want to copy the options passed in while determining publishing op…

abrandoned:
…tions

allows `in_options` to be frozen and not get a mutation when passed through

@mmmries


ztoolson:
:shipit:

mmmries:
Thanks @abrandoned and @ztoolson for getting this resolved

Issue #24, opened 2015-09-29
Cancel Subscriptions For Graceful Shutdown


mmmries:
Right now when you kill a process in `poll` mode, we stop polling for new messages and wait for the worker pool to empty out before exiting.

We should support this same behavior for `subscribe` mode by cancelling the subscriptions and waiting a few seconds for the worker pool to empty out before we actually shut down the connection and exit.

mmmries:
This was resolved in #30

PR #30, opened 2015-10-14
Graceful shutdown

mmmries:
This addresses #24 

This is an attempt to fix the problem where currently when you send a `SIGTERM` to your action_subscriber process in `subscribe` mode, it will close the connection immediately. This means that any events sitting in the local queue that were fetched without acknowledgement or any events that have been acknowledged and are sitting in the threadpool will be lost.

The fix is to keep class instance variables of all the consumers that we started up when the process started. Then when we catch a SIGTERM message we tell all those consumers to `cancel`. This informs RabbitMQ to cancel our subscription for new messages, but leaves the connection open so we can continue acknowledging and working on the messages we already received.

Then we go into a wait loop which checks the threadpool and waits for it to empty out (or gives up after 1000 seconds)
> Note: 1000 seconds was the current timeout for our graceful shutdown in `pop` mode, but it seems very excessive to me. I think we should reduce this to a configurable timeout which defaults to 60 seconds

jRuby output look like this during shutdown:
```
^CStopping server...
cancelling consumers for BaconSubscriber
finished cancelling consumers
waiting for threadpool to empty (2)
waiting for threadpool to empty (1)
threadpool empty. Shutting down
```

MRI output looks like this during shutdown:
```
^CStopping server...
cancelling consumers for BaconSubscriber
finished cancelling consumers
waiting for threadpool to empty (2)
waiting for threadpool to empty (2)
threadpool empty. Shutting down
```

/cc @film42 @quixoten @abrandoned @brettallred

### DIFF
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -41,14 +41,9 @@ module ActionSubscriber
   end
 
   if ::RUBY_PLATFORM == "java"
-    require "java"
-    class ShutdownHook
-      include java.lang.Runnable
-      def run
-        ::ActionSubscriber::Babou.stop_server!
-      end
+    at_exit do
+      ::ActionSubscriber::Babou.stop_server!
     end
-    java.lang.Runtime.getRuntime.addShutdownHook(java.lang.Thread.new(ShutdownHook.new))
   else
     [:INT, :QUIT, :TERM].each do |signal|
       trap(signal) do
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -47,12 +47,10 @@ module ActionSubscriber
 
       # Going to wait until the thread pool drains or we wait for 1000 seconds
       # Only waiting for shut down in pop mode
-      if ::ActionSubscriber::Babou.pop?
-        while ::ActionSubscriber::Threadpool.pool.busy_size > 0 && wait_loops < 1000
-          Thread.pass
-          wait_loops = wait_loops + 1
-          sleep 1
-        end
+      while ::ActionSubscriber::Threadpool.pool.busy_size > 0 && wait_loops < 1000
+        Thread.pass
+        wait_loops = wait_loops + 1
+        sleep 1
       end
 
       exit 0
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -11,6 +11,7 @@ module ActionSubscriber
     class_option :mode
     class_option :host
     class_option :hosts
+    class_option :prefetch, :type => :numeric, :desc => "how many messages to hold in the local queue in subscribe mode"
     class_option :pop_interval, :type => :numeric, :desc => "how long to wait between asking for messages (in milliseconds)"
     class_option :port, :type => :numeric
     class_option :threadpool_size, :type => :numeric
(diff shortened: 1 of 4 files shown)


bin/action_subscriber
```diff
@@ -11,6 +11,7 @@ module ActionSubscriber
     class_option :mode
     class_option :host
     class_option :hosts
+    class_option :prefetch, :type => :numeric, :desc => "how many messages to hold in the local queue in subscribe mode"
```
mmmries: I added a CLI option to set the prefetch just because it was convenient to my local testing, I can remove this if people think it should be in a different PR

film42: This is a feature I would expect to be in place. I think it's a simple enough change to include in this PR. :+1:

lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -26,8 +35,10 @@ def auto_pop!
 
       def auto_subscribe!
         queues.each do |queue|
-          queue.channel.prefetch(::ActionSubscriber.config.prefetch) if acknowledge_messages?
-          queue.subscribe(queue_subscription_options) do |delivery_info, properties, encoded_payload|
+          channel = queue.channel
+          channel.prefetch(::ActionSubscriber.config.prefetch) if acknowledge_messages?
+          consumer = ::Bunny::Consumer.new(channel, queue, channel.generate_consumer_tag, !acknowledge_messages?)
```
film42: This was interesting to read about. TIL more about rabbitmq.

film42:
:shipit:

mmmries:
I guess we can re-address the `1000 second timeout` in a future PR. I'll open a separate issue for that.

liveh2o:
The timeout seems excessive, but so does forcing everyone to use the same timeout. I say we make it configurable with a sensible default.

In the meantime, :shipit:

bin/action_subscriber
```diff
@@ -39,23 +40,16 @@ module ActionSubscriber
     end
   end
 
-  [:INT, :QUIT, :TERM].each do |signal|
-    trap(signal) do
-      puts "Stopping server..."
-      wait_loops = 0
+  if ::RUBY_PLATFORM == "java"
```
abrandoned: not understanding the split here .... in my experience jruby will catch TERM, QUIT, INT in most cases ..... of course if `stop_server!` is reentrant then we can register both instead of a separate line per platform

mmmries: if I just use `at_exit` then MRI throws an error about "`sleep': Interrupt"
* this doesn't actually cause an error, but seemed like it would be alarming to most users. I tried several workarounds like trying to wait for the sleep to finish and exiting the main thread of execution before I finished shutting down, but I couldn't get it to work cleanly

using the `trap(signal)` under jRuby causes it to hang during shutdown.
* I think this is because trapping these signals overrides some JVM stuff that is used for its own graceful shutdown

mmmries:
I published an `rc0` version from this branch: https://rubygems.org/gems/action_subscriber/versions/1.3.0.rc0

I'm going to manually pull that into one of our production applications and run it up to production just to be extra careful that this change doesn't break anything.

mmmries:
This has been running in production for several days and went through some serious inspection during that time (manually verifying, multiple restarts, etc).

I think this is good to go, but want to check in with the group before I merge and release as `1.3.0`?

PR #32, opened 2015-10-17
Exponential backoff

mmmries:
This is a rough cut at having an exponential backoff included in the `at_least_once!` mode. This first cut will retry messages on the following schedule:

<table>
<thead><tr><th>attempt</th><th>delay</th><th>total</th></tr></thead>
<tbody>
  <tr><td>1</td><td>0</td><td>0</td></tr>
  <tr><td>2</td><td>100ms</td><td>100ms</td></tr>
  <tr><td>3</td><td>500ms</td><td>600ms</td></tr>
  <tr><td>4</td><td>2.5s</td><td>3.1s</td></tr>
  <tr><td>5</td><td>12.5s</td><td>15.6s</td></tr>
  <tr><td>6</td><td>62.5s</td><td>78.1s</td></tr>
  <tr><td>7</td><td>5.1min</td><td>6.5m</td></tr>
  <tr><td>8</td><td>26min</td><td>32.5min</td></tr>
  <tr><td>9</td><td>2.1hr</td><td>2.7hr</td></tr>
  <tr><td>10</td><td>10.9hr</td><td>13.6hr</td></tr>
</tbody>
</table>

I don't like the names that are here yet, but I like the general idea of having 1 exchange and 1 queue (lazily created) per delay period per exchange.


film42:
Woooo hoooo! :+1:

lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -15,6 +15,7 @@ def auto_pop!
               :content_type => properties[:content_type],
               :delivery_tag => delivery_info.delivery_tag,
               :exchange => delivery_info.exchange,
+              :headers => properties.headers,
```
mmmries: I should probably add an integration test for both `subscribe` and `pop` modes to make sure that the headers always come through correctly and have a consistent type (ie `Hash[String => String]`)

lib/action_subscriber/message_retry.rb
```diff
@@ -0,0 +1,38 @@
...
+          :routing_key => env.routing_key,
+          :expiration => ttl
+        )
+      end
+    end
+
+    # Private Implementation
+    def self.calculate_ttl(env)
+      death = env.headers.fetch("x-death", [{"original-expiration" => "20"}]).first
+      next_ttl = death["original-expiration"].to_s.to_i * 5
```
mmmries: this `to_s` is because march_hare is currently passing the list of headers as `JavaCharacterArray`s. So I had to convert to a ruby string before calling `to_i`, but really that conversion should be happening when I build the `Middleware::Env` object.

lib/action_subscriber/message_retry.rb
```diff
@@ -0,0 +1,38 @@
...
+      return false if next_ttl > 86_400_000
+      next_ttl
+    end
+
+    def self.with_exchange(env, ttl)
+      retry_name = "#{env.exchange}_retry_#{ttl}"
+      channel = RabbitConnection.subscriber_connection.create_channel
+      begin
+        channel.confirm_select
```
mmmries: we could potentially create all these topic exchanges and queues at startup time, but since message retries aren't designed to be part of the "hot path" I figured it was okay to just lazily create them when a retry occurs.

lib/action_subscriber/message_retry.rb
```diff
@@ -0,0 +1,38 @@
...
+    # Private Implementation
+    def self.calculate_ttl(env)
+      death = env.headers.fetch("x-death", [{"original-expiration" => "20"}]).first
+      next_ttl = death["original-expiration"].to_s.to_i * 5
+      return false if next_ttl > 86_400_000
+      next_ttl
+    end
+
+    def self.with_exchange(env, ttl)
+      retry_name = "#{env.exchange}_retry_#{ttl}"
```
mmmries: one clarification here, the reason we have to generate an exchange and queue pair for each separate timeout period and exchange name is because the queue can only specify one dead letter exchange (where the message will published when it expires) and messages only get expired when they are at the head of a queue. So if there were a long TTL message ahead of a shorter TTL message, the shorter one would not expire until the longer one expired.

So keeping a unique queue for each period is necessary and keeping a unique queue for each exchange is necessary. I don't think that having extra exchanges and queues will be a significant overhead, but we should probably test it to make sure that is the case.

mmmries:
@quixoten @abrandoned @brianstien @liveh2o @localshred @brettallred 

This could definitely use some review especially around:
* naming
* reliability
* discoverability

lib/action_subscriber/message_retry.rb
quixoten: Could a single retry exchange be used if each original queue has a retry queue in the retry exchange for every ttl? It might be easier introspect retries (through the web UI for example) if they're all happening under a single exchange, but maybe it's sixes?

mmmries: @quixoten if we mirror the set of queue names we would end up with on extra exchange (per exchange that our subscribers listen to) and one queue per subscription queue and timeout level. This would create a total of 10x the number of queues we currently have in production.

I really like the idea of being able to see which queues are getting retry messages, but I'm not sure how much overhead is involved with having that many queues exist in the rabbit cluster. Perhaps I can come up with some sort of benchmark to measure the impact.

mmmries: Checking into this a bit further it doesn't look like queues cost much in terms of CPU or RAM. In our production system we are using about 19MB per rabbitmq instance for queues. So if we 10x the number of queues in the system that would take up about 190MB of RAM.

Each connection / channel / consumer takes up a lot more resources than the queues themselves. And since these queues will just be used to hold messages while we wait for them to expire we won't actually have any channels or consumers connected to them.

screen shot 2015-10-20 at 9 54 40 am

quixoten: When I made my initial comment I was thinking that you would be making `retry_timeouts * number_of_original_queues` retry queues either way, but as it's currently coded it's really `retry_timeouts * number_of_original_exchanges`, right?

lib/action_subscriber/bunny/subscriber.rb
mmmries: I added an integration test for headers

mmmries:
@quixoten I made the change so that we setup retry queues that are named after the original queue. I like the extra visibility that this provides.

screen shot 2015-10-20 at 4 25 37 pm

mmmries:
Looks like jRuby is running into an issue where ther `x-death` header isn't populating correctly. That's why we have a `_retry_0` queue. I'll dig into that before trying to move this forward

lib/action_subscriber/message_retry.rb
mmmries: yes, the first implementation was siginificantly less total queues, but it
looks like the size of the messages themselves dwarf the overhead of the
queues (unless your messages are EXTREMELY small)

So having better visibility seems like it would be worth the extra queues

On Tue, Oct 20, 2015 at 3:53 PM, Devin Christensen <notifications@github.com

mmmries:
It turns out that `bunny` was parsing the `x-death` header for us so when we access the header we get back an array with hashes inside of it, but `march_hare` is returning the string version like this:
`{reason=expired, original-expiration=100, count=1, exchange=events_retry_100, time=Tue Oct 20 21:18:04 MDT 2015, routing-keys=[gorby_puff.grumpy], queue=alice.gorby_puff.grumpy_retry_100}]`

:rage1: :rage2: :rage3: Good thing rabbitmq didn't just publish some JSON :rage1: :rage2: :rage3:

I'm wondering if I should try to write a parser to pull the data out manually? I could open a PR in the `march_hare` gem to match Bunny's behavior, but then we would have to lock down to only use the most recent versions of `march_hare`...

quixoten:
Would it be possible to tie into the amqp-protocol gem directly? I'm just
derping around on my phone but it looks like that's where the decoding of
the headers happen. Or are we trying to avoid adding any extra dependencies?

mmmries:
@quixoten I can't find anything related to parsing `x-death` headers in the `amq-protocol` gem. What did you find?

quixoten:
I was looking [here](https://github.com/ruby-amqp/amq-protocol/blob/1767353d9a8d60dd50934b45d57b3f8b7a91b3c3/lib/amq/protocol/frame.rb#L59-L71), but I'm still rather unfamiliar with the code.

lib/action_subscriber/message_retry.rb
```diff
@@ -0,0 +1,51 @@
...
+    # Private Implementation
+    def self.calculate_ttl(env)
+      last_ttl = get_last_ttl_or_default(env.headers["x-death"], 20)
+      next_ttl = last_ttl * 5
+      return false if next_ttl > 86_400_000
+      next_ttl
+    end
+
+    def self.get_last_ttl_or_default(x_death, default)
+      return default unless x_death
```
mmmries: I implemented the parsing logic here locally by only handling the simple cases (ie. this doesn't parse all the keys and values out of the header, but it covers the ones we care about)

quixoten:
Yeah, the decoding in amqp-protocol seems to be occurring at a lower level. `{reason=expired, original-expiration=100, count=1, exchange=events_retry_100, time=Tue Oct 20 21:18:04 MDT 2015, routing-keys=[gorby_puff.grumpy], queue=alice.gorby_puff.grumpy_retry_100}]` looks like post-decoded data, almost like it's a pretty printed form of the decoded data.

mmmries:
Okay. I just realized last night that all of this parsing stuff is not necessary. I was only using the information to decide how long to set the TTL for the next retry, but I can easily track that in my own header rather than using the existing `x-death` header.

Just pushed a commit that changes the `MessageRetry` to append a `as-attempt` header to all retries. It then uses the attempt number to determine the TTL. I also changed it to set the TTL on the queue rather than on each message.

lib/action_subscriber/base.rb
```diff
@@ -80,7 +80,8 @@ def _at_least_once_filter
       yield
       acknowledge
     rescue => error
-      reject
+      ::ActionSubscriber::MessageRetry.redeliver_message_with_backoff(env)
```
mmmries: I wanted to point out here, that we are publishing the message retries with `publisher_confirms` on, so this pattern here means that if our code crashes between publishing the retry and when we acknowledge we could end up duplicating the message.

I'm not very worried about this because if people are turning on `at_least_once!` I assume they are okay with making their subscriber idempotent, but just wanted to point it out in case someone thinks of something I'm missing

mmmries:
I'm not happy with this PR. To sum up the changes:

* bunny and march hare now provide the `queue` (String) and `headers` (Hash[String => String])` to the env object for each subscriber event
* when rescuing errors for `at_least_once!` mode we call `MessageRetry.redeliver_message_with_backoff(env) && acknowledge` instead of the simplistic `reject`
* `MessageRetry` is idempotent and re-entrant (thanks RabbitMQ) so it creates the retry exchanges and retry queues lazily (this is not the happy pathy so we're not super concerned about performance here)
* the extra exchanges and queues in RabbitMQ don't appear to cost much in terms of production hardware (an extra 200MB of RAM for the extra queues + the RAM to hold the messages that need to get re-delivered, rabbitmq pages these disk pretty efficiently)

I would like to release a 1.4.0.rc0 version of the gem and try it out in one or two subscribers in our production environment at work to watch for memory leaks or any other issues that may come up.

Any objections? @quixoten @abrandoned @liveh2o @localshred @brettallred

mmmries:
BTW our production rabbit machines right now hover around 250MB used, and during the highest loads (background aggregations) they spike to ~300MB which represents ~10k messages in their queues. They are currently set for a memory usage high watermark of 2.3GB so we have a headroom of about 2GB which would be on the order of ~400k retry messages hanging around.

mmmries:
I released `1.4.0.pre0` version of the gem so I can do a test in one of our production applications.

mmmries:
This has been deployed to a production system at MX and has been running for over 24 hours. I think we should go ahead and release `1.4.0` since this is an opt-in feature and we have some solid evidence that there are no regressions.

/cc @abrandoned @film42 @brettallred @localshred

film42:
bzbqxuo

:shipit:

PR #33, opened 2015-10-20
Warn when invalid configuration.mode specified

newellista:
Currently, if the ::ActionSubscriber.configuration.mode is invalid, the
application simply fails quietly.

This change will print an error to the console, and hint at what needs
to be changed.

@mmmries @brianstien

### DIFF
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -35,6 +35,8 @@ module ActionSubscriber
         ::ActionSubscriber::Babou.auto_pop!
       when /subscribe/i then
         ::ActionSubscriber::Babou.start_subscribers
+      else
+        puts "ActionSubscriber.configuration.mode must be 'pop' or 'subscribe'.  Currently set to '#{::ActionSubscriber.configuration.mode}'"
       end
     end
   end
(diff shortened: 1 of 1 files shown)


mmmries:
Nice catch. That is a pretty big oversight.
I vote we :shipit:

bin/action_subscriber
```diff
@@ -35,6 +35,8 @@ module ActionSubscriber
         ::ActionSubscriber::Babou.auto_pop!
       when /subscribe/i then
         ::ActionSubscriber::Babou.start_subscribers
+      else
+        puts "ActionSubscriber.configuration.mode must be 'pop' or 'subscribe'.  Currently set to '#{::ActionSubscriber.configuration.mode}'"
```
film42: We should probably `fail` here instead of logging to stdout. I would find that more helpful than a log message. #MyOpinion

newellista:
I agree with @film42.  Changed `puts` to `fail`.

mmmries:
it will be nice to have a bad exit code when we hit that fail :+1:
Still a :shipit: for me

film42:
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

PR #34, opened 2015-11-06
Routes

mmmries:
We have plans to make it possible to have per-action acknowledgements and per-action exchange settings. This PR is a prelude to those features.

> Make the change easy...then make the easy change.
~Kent Beck [source](https://twitter.com/kentbeck/status/250733358307500032)

This PR introduces a `RouteSet` object which holds a bunch of `Route` objects and can then setup queues and start subscribing to those routes. This shrinks the public interface of `ActiveRecord::Base` and shifts some of the responsibility to a new object.

In future PR's we will introduce a new way to "draw routes" and make `DefaultRouting` optional.

/cc @quixoten @localshred @abrandoned @brianstien 

_PS: I'm sorry for the size of this PR. I tried to find a smaller cut at this that made any significant progress and this was the smallest chunk I could find._

### DIFF
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -77,7 +77,7 @@ module ActionSubscriber
   end
 
   def self.stop_subscribers!
-    route_set.cancel_consumers
+    route_set.cancel_consumers!
   end
 
   ##
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -80,6 +80,10 @@ module ActionSubscriber
     print_subscriptions
   end
 
+  def self.stop_subscribers!
+    @route_set.cancel_consumers! if @route_set
+  end
+
   ##
   # Class aliases
   #
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -22,6 +22,8 @@ require "action_subscriber/bunny/subscriber"
 require "action_subscriber/march_hare/subscriber"
 require "action_subscriber/babou"
 require "action_subscriber/publisher"
+require "action_subscriber/route"
+require "action_subscriber/route_set"
 require "action_subscriber/threadpool"
 require "action_subscriber/base"
 
@@ -35,18 +37,14 @@ module ActionSubscriber
   #
   def self.auto_pop!
     return if ::ActionSubscriber::Threadpool.busy?
-    ::ActionSubscriber::Base.inherited_classes.each do |klass|
-      klass.auto_pop!
-    end
+    @route_set.auto_pop!
   end
 
   # Loop over all subscribers and register each as
   # a subscriber.
   #
   def self.auto_subscribe!
-    ::ActionSubscriber::Base.inherited_classes.each do |klass|
-      klass.auto_subscribe!
-    end
+    @route_set.auto_subscribe!
   end
 
   def self.configuration
@@ -62,8 +60,10 @@ module ActionSubscriber
   end
 
   def self.setup_queues!
-    ::ActionSubscriber::Base.inherited_classes.each do |klass|
-      klass.setup_queues!
+    @route_set ||= begin
+      route_set = RouteSet.new(routes)
+      route_set.setup_queues!
+      route_set
     end
   end
 
@@ -91,6 +91,17 @@ module ActionSubscriber
   config
 
   ::ActiveSupport.run_load_hooks(:action_subscriber, Base)
+
+  ##
+  # Private Implementation
+  #
+  def self.routes
+    ::ActionSubscriber::Base.inherited_classes.flat_map do |klass|
+      klass.routes
+    end
+  end
+  private_class_method :routes
+
 end
 
 require "action_subscriber/railtie" if defined?(Rails)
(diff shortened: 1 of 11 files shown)


mmmries:
Because this PR is on the large side and touches a lot of internals I thought a visual might help with reviewing. The essence of the changes in this PR are about who manages the "queues" that receive messages from RabbitMQ.

### Before This PR
screen shot 2015-11-06 at 8 18 50 am

Currently when `ActionSubscriber.setup_queues!` is called it proxies that call to each of the Subscriber classes which then create and manage their own list of queues and they mixin the `Bunny::Subscriber` or `MarchHare::Subscriber` to manage those queues and get messages from them.

### In This PR
screen shot 2015-11-06 at 8 22 00 am

The Subscriber classes are now responsible for creating a list of `Route`s which contain all the settings for a queue, these are then passed to a `RouteSet` object (don't like this name) which creates and manages the queues. Having an "at-rest" definition for all the queues is what opens the door for us to introduce additional routing functionality. We can now generate that list of `Route`s by any method we want which should make it easier to have some actions subscribe to one exchange and other actions subscribe to a different exchange.

lib/action_subscriber.rb
```diff
@@ -80,6 +80,10 @@ def self.start_subscribers
     print_subscriptions
   end
 
+  def self.stop_subscribers!
+    @route_set.cancel_consumers! if @route_set
```
film42: Why is there are guard here for `@route_set`, but not other places in this code?

mmmries: this code path is called during a graceful shutdown so I'm just guarding against the race condition of someone starting up the subscriber and killing it before ActionSubscriber has actually finished starting up. I didn't want to throw a false warning, but maybe I should add guard to the other places we access the route set?

liveh2o: I wouldn't be accessing the class-instance variable. If you make a method for it, you can incorporate the guard clause in it.

lib/action_subscriber.rb
```diff
@@ -62,8 +60,10 @@ def self.print_subscriptions
   end
 
   def self.setup_queues!
-    ::ActionSubscriber::Base.inherited_classes.each do |klass|
-      klass.setup_queues!
+    @route_set ||= begin
```
liveh2o: If you split initializing the route set out from the setup queues step, you won't need to add any guard clauses. I'd follow the same pattern as initializing the config in this class.

lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -15,8 +14,8 @@ def auto_pop!
         # of times we will pop each time we poll the broker
         times_to_pop = [::ActionSubscriber::Threadpool.ready_size, ::ActionSubscriber.config.times_to_pop].min
         times_to_pop.times do
-          queues.each do |queue|
-            delivery_info, properties, encoded_payload = queue.pop(queue_subscription_options)
```
liveh2o: Where was `queue_subscription_options` even coming from?

lib/action_subscriber/default_routing.rb
```diff
@@ -1,25 +1,21 @@
...
-    def setup_queues!
-      exchange_names.each do |exchange_name|
-        subscribable_methods.each do |method_name|
-          queues << setup_queue!(method_name, exchange_name)
+    def routes
+      @routes ||= begin
+        routes = []
+        exchange_names.each do |exchange_name|
+          subscribable_methods.each do |method_name|
+            routes << ActionSubscriber::Route.new({
```
liveh2o: I'm curious as to why these keys don't shadow the methods that set them? Not a big deal. I'm just thinking about discoverability.

lib/action_subscriber/route_set.rb
```diff
@@ -0,0 +1,35 @@
+module ActionSubscriber
+  class RouteSet
+    if ::RUBY_PLATFORM == "java"
```
liveh2o: I like that we've limited the impact of this platform switch.

lib/action_subscriber/bunny/subscriber.rb
mmmries: It was coming from the `ActionSubscriber::DSL` module which gets mixed into every `ActionSubscriber::Base` class. That is where we currently manage whether a subscriber is going to use message acknowledgements, but now we are managing that in the route so I just moved the method from the module into the `Route` object.

lib/action_subscriber/default_routing.rb
mmmries: The naming of these keys is what I think makes the most sense for the project moving forward. I'm planning on deprecating, renaming or removing the method names seen here.

lib/action_subscriber.rb
mmmries: I've pulled out the `route_set` to its own method that makes sure it is initialized to go along with your comments here. I think there is a future cut here to separate the list of routes from the state of  setting up consumers and declaring those queues. But for now this is a definite improvement :+1:

film42:
I've read through this a few times. Looks good to me. :shipit:

Issue #25, opened 2015-09-29
Add Logging


mmmries:
Not having any subscriber logs has been a pain point for us in the past. We have statsd metrics around the events which alleviates some of the pain, but it would be really helpful to also keep a local log file for each subscriber.

PR #35, opened 2015-11-10
Explicit routes

mmmries:
:warning: :warning: use this [diff link](https://github.com/mxenabled/action_subscriber/compare/routes...explicit_routes) until #34 is merged :warning: :warning:

This is a new feature that deprecates the old `DefaultRouting` mixin for the subscribers. Instead of having the library discover subscribers and infer their routes, we are going to let the application draw their own routes similar to Rails' `config/routes.rb` file.

The `routing.md` file has some examples and more explanation, but here is a simple example to whet your appetite.
```ruby
ActionSubscriber.draw_routes do
  default_routes_for ::UserSubscriber
  route ::AccountSubscriber, :created, :publisher => :abacus, :exchange => :events
  route ::AccountSubscriber, :refresh, :publisher => "*", :exchange => :actions
end
```

/cc @abrandoned @localshred @liveh2o @brettallred

### DIFF
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -77,7 +77,7 @@ module ActionSubscriber
   end
 
   def self.stop_subscribers!
-    route_set.cancel_consumers
+    route_set.cancel_consumers!
   end
 
   ##
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -62,7 +62,16 @@ module ActionSubscriber
   end
 
   def self.print_subscriptions
-    ::ActionSubscriber::Base.print_subscriptions
+    puts configuration.inspect
+    route_set.routes.group_by(&:subscriber).each do |subscriber, routes|
+      puts subscriber.name
+      routes.each do |route|
+        puts "  -- method: #{route.action}"
+        puts "    --    exchange: #{route.exchange}"
+        puts "    --       queue: #{route.queue}"
+        puts "    -- routing_key: #{route.routing_key}"
+      end
+    end
   end
 
   def self.setup_queues!
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -80,6 +80,10 @@ module ActionSubscriber
     print_subscriptions
   end
 
+  def self.stop_subscribers!
+    @route_set.cancel_consumers! if @route_set
+  end
+
   ##
   # Class aliases
   #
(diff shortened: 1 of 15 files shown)


mmmries:
I released `action_subscriber-1.5.0.pr0` so we can try this out (https://rubygems.org/gems/action_subscriber/versions/1.5.0.pre0-java).

lib/action_subscriber.rb
```diff
@@ -57,14 +56,26 @@ def self.configure
...
   end
 
+  def self.draw_routes(&block)
+    routes = Router.draw_routes(&block)
+    @route_set = RouteSet.new(routes)
+  end
+
   def self.print_subscriptions
-    ::ActionSubscriber::Base.print_subscriptions
+    puts configuration.inspect
```
quixoten: Should these puts be converted to a logger ?

ryanbjones:
The logic on all of this looks really sound and clean :+1: I'll let someone with more domain knowledge check for any edge cases and put their stamp of approval on it.

lib/action_subscriber.rb
mmmries: @quixoten I am planning on switching all `puts` to `Logger`s when I tackle #25, but right now ActionSubscriber doesn't handle any logging, just standard out

quixoten:
All the specs passed on my box! Code looks good. :shipit:

film42:
@mmmries Does this mean we'll need to add `default_routes_for  ::Model` for all subscribers for every app before we roll this out?

quixoten:
@film42:

> the old behavior of discovering and inferring routes for subscribers will be supported until the 2.0 version of ActionSubscriber. When no routes are drawn before calling `setup_queues!` we will print a deprecation warning and infer the routes for you.

mmmries:
@quixoten @film42  :+1: the fallback behavior with deprecation message is here: https://github.com/mxenabled/action_subscriber/blob/47672e36bbbbbf08b800b0b4e2197279d55efc30/lib/action_subscriber.rb#L113-L119

mmmries:
basically if you haven't drawn your own routes, then I give you a deprecation warning and draw them for you.

film42:
:+1: :+1: :+1: :+1: :+1: :shipit:

mmmries:
I've picked up the pre-release version of this gem into one of our internal projects at MX. It is on its way to production and after giving it a few days there to watch for stability I'll release this.

mmmries:
This ran in production overnight and processed several million messages. Everything still looks good so I'm going to go ahead and release a public `1.5.0` version

PR #38, opened 2015-12-04
Prefetch per route

mmmries:
We have a case in one of our subscribers where the average time to process one job type is MUCH longer than other types (30sec avg vs 10ms avg). Because of this disparity we would like to have that route prefetch much fewer messages than the other routes. This will help to avoid that queue getting backed up waiting for acknowledgements on a busy server.

This new feature just allows us to set a `prefetch` on a specific route when we are manually drawing the routes. It results in a set of channels like this:
screen shot 2015-12-03 at 4 58 37 pm

/cc @ztoolson @liveh2o @brianstien @abrandoned @quixoten


ztoolson:
Changes look good. :shipit:

PR #39, opened 2015-12-08
use ActionSubscriber::Publisher for all publishing in tests

mmmries:
This changes our integration tests to use `ActionSubscriber::Publisher.publish` rather than manually grabbing connection objects and doing publishing. No behavioral changes here, just a refactoring to make the tests easier to understand.

/cc @film42 @localshred @quixoten @liveh2o


quixoten:
:shipit:

PR #40, opened 2015-12-15
Draw routes for tests

mmmries:
As we are working towards deprecating inferred routes I wanted to have the test suite draw explicit routes. I also added a new test which specifically doesn't draw its own routes to make sure we don't break this deprecated behavior until we are ready to cut 2.0.

This also makes it a lot easier for me to make the next change of supporting multiple threadpools.

/cc @quixoten @localshred @abrandoned @liveh2o


camkidman:
Seems all good to me, but you might want another set of :eyes: as well!

quixoten:
:shipit:

PR #41, opened 2015-12-15
Multiple threadpools

### LINKED ISSUE #37: Multiple threadpools
I would really like to deprecate the `--allow-low-priority-queues` option. Running two separate subscriber processes is a pain when what we really want is just to avoid having the low priority work take up all the threads in the threadpool.

I propose that we make it possible to assign a `route` to a named threadpool. This would allow us to setup a `low_priority` threadpool with just a few threads and connect all the `*_low` subscriptions to that threadpool while we keep the rest of the subscriptions connected to the main threadpool.

`config/action_subscriber.yml`
```yaml
production:
  threadpool_size: 10 #the simpler form (will still be supported)
  threadpool_sizes:
    default: 10
    low_priority: 5
```

`config/initializers/action_subscriber.rb`
```ruby
ActionSubscriber.draw_routes do
  route ::ClientSnapshotSubscriber, :created, :acknowledgements => true, :publisher => :amigo
  route ::AggregationSubscriber, :completed_low, :threadpool => :low_priority, :acknowledgements => true, :publisher => :grunt
end
```

Or perhaps we could skip the yaml configuration and just set things up in the route definitions like:

```ruby
low_priority_threadpool = ::ActionSubscriber::Threadpool.new(5)

  route ::ClientSnapshotSubscriber, :created, :acknowledgements => true, :publisher => :amigo
  route ::AggregationSubscriber, :completed_low, :threadpool => low_priority_threadpool, :acknowledgements => true, :publisher => :grunt
end
```

/cc @abrandoned @localshred @liveh2o @brianstien

mmmries:
:warning: :warning: __use [this diff link](https://github.com/mxenabled/action_subscriber/compare/draw_routes_for_tests...multiple_threadpools)__. This PR contains the changes from #40 as well.

This fixes #37. It allows an application to specify multiple threadpools of varying sizes and then direct different routes into those threadpools. This means that if you have routes with disparate latencies you can keep the fast jobs separate from the slow jobs. This also opens the door for us to deprecate the `--allow-low-priority-methods` flag. In the `2.0` release I would like to remove that feature and delete all of the code that is branching around that flag.

/cc @abrandoned @brettallred @liveh2o @localshred @quixoten

### DIFF
--- a/lib/action_subscriber/threadpool.rb
+++ b/lib/action_subscriber/threadpool.rb
@@ -4,17 +4,29 @@ module ActionSubscriber
     # Class Methods
     #
     def self.busy?
-      (pool.pool_size == pool.busy_size)
+      pools.any? do |_pool_name, pool|
+        pool.pool_size == pool.busy_size
+      end
     end
 
-    def self.perform_async(*args)
-      self.pool.async.perform(*args)
-    end
-
-    def self.pool
-      @pool ||= ::Lifeguard::InfiniteThreadpool.new(
-        :pool_size => ::ActionSubscriber.config.threadpool_size
+    def self.new_pool(name, pool_size = nil)
+      fail ArgumentError, "#{name} already exists as a threadpool" if pools.key?(name)
+      pool_size ||= ::ActionSubscriber.config.threadpool_size
+      pools[name] = ::Lifeguard::InfiniteThreadpool.new(
+        :pool_size => pool_size
       )
+    end
+
+    def self.pool(which_pool = :default)
+      pools[which_pool]
+    end
+
+    def self.pools
+      @pools ||= {
+        :default => ::Lifeguard::InfiniteThreadpool.new(
+          :pool_size => ::ActionSubscriber.config.threadpool_size
+        )
+      }
     end
 
     def self.ready?
@@ -22,8 +34,9 @@ module ActionSubscriber
     end
 
     def self.ready_size
-      ready_size = pool.pool_size - pool.busy_size
-      return ready_size >= 0 ? ready_size : 0
+      pools.inject(0) do |total_ready, (_pool_name, pool)|
+        total_ready + [0, pool.pool_size - pool.busy_size].max
+      end
     end
   end
 end
--- a/spec/integration/multiple_threadpools_spec.rb
+++ b/spec/integration/multiple_threadpools_spec.rb
@@ -1,0 +1,29 @@
+class DifferentThreadpoolsSubscriber < ActionSubscriber::Base
+  def one
+    $messages << payload
+  end
+
+  def two
+    $messages << payload
+  end
+end
+
+describe "Separate Threadpools for Different Message", :integration => true do
+  let(:draw_routes) do
+    low_priority_threadpool = ::ActionSubscriber::Threadpool.new_pool(:low_priority, 1)
+    ::ActionSubscriber.draw_routes do
+      route DifferentThreadpoolsSubscriber, :one
+      route DifferentThreadpoolsSubscriber, :two, :threadpool => low_priority_threadpool
+    end
+  end
+
+  it "processes messages in separate threadpools based on the routes" do
+    ::ActionSubscriber.auto_subscribe!
+    ::ActionSubscriber::Publisher.publish("different_threadpools.one", "ONE", "events")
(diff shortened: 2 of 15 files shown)


lib/action_subscriber/threadpool.rb
```diff
@@ -4,17 +4,29 @@ class Threadpool
...
-      @pool ||= ::Lifeguard::InfiniteThreadpool.new(
-        :pool_size => ::ActionSubscriber.config.threadpool_size
-      )
+    def self.pool(which_pool = :default)
+      pools[which_pool]
+    end
+
+    def self.pools
+      @pools ||= {
+        :default => ::Lifeguard::InfiniteThreadpool.new(
```
quixoten: It feels like maybe this should also be using `self.new_pool` to create the pool so there's not to paths for the same logic. Maybe something like:

```ruby
def self.new_pool
  # same as above
end

def self.pools; @pools ||= {}; end

self.new_pool(:default) unless self.pools.key?(:default)
```

I'd also consider renaming `new_pool` to `add_pool` or just `add`.

spec/integration/multiple_threadpools_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it "processes messages in separate threadpools based on the routes" do
+    ::ActionSubscriber.auto_subscribe!
+    ::ActionSubscriber::Publisher.publish("different_threadpools.one", "ONE", "events")
+    ::ActionSubscriber::Publisher.publish("different_threadpools.two", "TWO", "events")
+
+    verify_expectation_within(2.0) do
+      expect($messages).to eq(Set.new(["ONE","TWO"]))
+    end
+  end
+end
```
quixoten: Would it be possible to test that the pool is of the set size? Maybe adding a `sleep` after each `$messages << payload` and then expecting only the first message to be added to the list?

mmmries: I tried writing the test that way a few times, but because the timing of publishing the message to rabbitmq and getting it back from rabbitmq is asynchronous and varies a lot between running locally vs running in travis I just couldn't get that test to run consistently.

That's why I changed it to just be a smoke test.

mmmries:
By the way I tested this locally by making a subscriber where one route takes 10 seconds per job and another route takes 0.1 seconds per job. When I kicked off a bunch of jobs the queues looked like this:

screen shot 2015-12-15 at 10 08 09 pm

screen shot 2015-12-15 at 10 08 13 pm

You can see that the "overcooked" queue is draining very slowly, but even though it has saturated its threadpool the "cooked" jobs drain almost immediately because they are in a separate threadpool. This is a manual version of the test that @quixoten suggested we automate.

lib/action_subscriber/threadpool.rb
mmmries: I tried rewriting this a few different ways (including your suggestion of calling `self.add_pool(:default)`, but I can't find anything I like.

The problem with calling `self.new_pool(:default)` as I'm defining this class is that the application calling my code has not yet had a chance to call `ActionSubscriber.config` to set the default pool size. Normally we don't initialize the pool until it first gets accessed which is after the user has done their configuration.

The duplication does bother me a little, but I haven't been able to find a better way so far. Usually when I find myself in this situation I fall back to:

> Duplication is far cheaper than the wrong abstraction
> ~ Sandi Metz

ryanbjones:
This looks good to me :+1:

mmmries:
Final call for comments on things we need to improve before rolling this out in a minor version release.

I'm happy to slow down for feedback, but if I'm not getting any feedback I'll move this forward.

lib/action_subscriber/threadpool.rb
```diff
@@ -4,17 +4,29 @@ class Threadpool
...
+      )
     end
 
-    def self.pool
-      @pool ||= ::Lifeguard::InfiniteThreadpool.new(
-        :pool_size => ::ActionSubscriber.config.threadpool_size
-      )
+    def self.pool(which_pool = :default)
+      pools[which_pool]
+    end
```
vintagepenguin: What do you think about the name `current_pool` instead of `which_pool`?  

`which_pool` kind of confused me when I was reading through this.

mmmries:
This has now been in an MX production system for a full day and it looks stable and performant. Time to release `1.7.0`

PR #43, opened 2015-12-23
Logging

### LINKED ISSUE #25: Add Logging

mmmries:
Fixes #25 

We have needed logging for a long time, so let's get this done. I lifted the `Logging` class from sidekiq and removed a bunch of stuff that I don't think we need (at least initially) like multiple custom formatters, re-opening logger objects when receiving `USR2` signals etc.

Then I changed most of the `puts` statements in `bin/action_subscriber`, `lib/action_subscriber.rb` and `lib/action_subscriber/babou.rb` to use the new logger. Finally I added a little bit of logging for when we receive messages from Rabbitmq, when we start processing those messages, when we finish processing them and when a message fails to process.

Then I tested it under both MRI and jRuby. Fixed an edge case where you can't use a `Logger` when being called from a `trap` context and I ended up with output looking like this:
screen shot 2015-12-23 at 11 00 18 am

/cc @film42 @abrandoned @localshred @liveh2o @brianstien

### DIFF
--- a/lib/action_subscriber/middleware/env.rb
+++ b/lib/action_subscriber/middleware/env.rb
@@ -1,3 +1,5 @@
+require "securerandom"
+
 module ActionSubscriber
   module Middleware
     class Env
@@ -30,7 +32,7 @@ module ActionSubscriber
         @encoded_payload = encoded_payload
         @exchange = properties.fetch(:exchange)
         @headers = properties.fetch(:headers) || {}
-        @message_id = properties.fetch(:message_id) || ::Random.new.bytes(3).unpack("H*")[0]
+        @message_id = properties.fetch(:message_id) || ::SecureRandom.hex(3)
         @queue = properties.fetch(:queue)
         @routing_key = properties.fetch(:routing_key)
         @subscriber = subscriber
--- a/lib/action_subscriber/middleware/env.rb
+++ b/lib/action_subscriber/middleware/env.rb
@@ -30,7 +30,7 @@ module ActionSubscriber
         @encoded_payload = encoded_payload
         @exchange = properties.fetch(:exchange)
         @headers = properties.fetch(:headers) || {}
-        @message_id = properties.fetch(:message_id)
+        @message_id = properties.fetch(:message_id) || ::Random.new.bytes(3).unpack("H*")[0]
         @queue = properties.fetch(:queue)
         @routing_key = properties.fetch(:routing_key)
         @subscriber = subscriber
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -6,8 +6,6 @@ require 'thor'
 
 module ActionSubscriber
   class CLI < ::Thor
-    include ::ActionSubscriber::Logging
-
     class_option :allow_low_priority_methods, :type => :boolean, :desc => "subscribe to low priority queues in addition to the normal queues", :default => false
     class_option :app, :default => "./config/environment.rb"
     class_option :mode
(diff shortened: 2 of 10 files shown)


lib/action_subscriber/middleware/env.rb
```diff
@@ -30,7 +30,7 @@ def initialize(subscriber, encoded_payload, properties)
         @encoded_payload = encoded_payload
         @exchange = properties.fetch(:exchange)
         @headers = properties.fetch(:headers) || {}
-        @message_id = properties.fetch(:message_id)
+        @message_id = properties.fetch(:message_id) || ::Random.new.bytes(3).unpack("H*")[0]
```
mmmries: Publishing with a message_id is optional (basically a custom header) but I wanted to use that ID if it was provided so this just generates a random `message_id` if nothing was provided by the publisher of the message.

mmmries: PS: I used `Random.new.bytes` instead of `SecureRandom.hex` because I'm assuming the overhead is less, but maybe it doesn't matter. Anyone know if there are concerns around running out of entropy or other concerns with `SecureRandom` vs `Random`?

mmmries:
BTW I cut a [prerelease version](https://rubygems.org/gems/action_subscriber/versions/1.7.0.pre0) of the gem with this branch and the `multiple_threadpools` branch in it so I can test these changes out  more thoroughly

lib/action_subscriber/middleware/env.rb
abrandoned: just for giggles I ran the benchmark ... SecureRandom.hex is 5x faster than just Random.new.bytes(10) (with no unpack executed) 

don't know what it is doing, but looks like SecureRandom is "better"

abrandoned:
big fan of this PR :shipit:

mmmries:
This is now in a production system at MX and the logs look good:

```
I, [2016-01-04T14:42:51.558000 #7219]  INFO -- : RECEIVED 361e5f from bunyan.abacus.account.updated
I, [2016-01-04T14:42:51.559000 #7219]  INFO -- : START 361e5f AccountSubscriber#updated
I, [2016-01-04T14:42:51.571000 #7219]  INFO -- : FINISHED 361e5f
I, [2016-01-04T14:42:51.590000 #7219]  INFO -- : RECEIVED c8aee6 from bunyan.amigo.user.changed
I, [2016-01-04T14:42:51.590000 #7219]  INFO -- : START c8aee6 UserSubscriber#changed
I, [2016-01-04T14:42:51.591000 #7219]  INFO -- : FINISHED c8aee6
I, [2016-01-04T14:42:51.603000 #7219]  INFO -- : RECEIVED 55ca50 from bunyan.persona.member.updated
I, [2016-01-04T14:42:51.604000 #7219]  INFO -- : START 55ca50 MemberSubscriber#updated
I, [2016-01-04T14:42:51.632000 #7219]  INFO -- : FINISHED 55ca50
I, [2016-01-04T14:42:51.662000 #7219]  INFO -- : RECEIVED a06cdf from bunyan.abacus.account.updated
```

The mean time to process messages does not seem to be affected at all (still ~10-25ms per job)

mmmries:
* rebased this against master (with the multiple threadpools merged)
* changed the `Random.bytes.unpack...` mess with just a `SecureRandom.hex(3)` call
* silence the logger during the test suite (it was very noisy)

PR #44, opened 2015-12-29
retry messages should be published directly to the queue

### LINKED ISSUE #42: Surprising Retry Behavior
In talking with @abrandoned about a similar system I realized that if you have two subscribers that are both listening to the same routing key, and one of them fails a job and retries it (with the new `at_least_once!` mode), the job would be re-published to both of the subscribers.

This seems like a very surprising behavior to me. I think each of those subscribers should be able to set themselves as `at_least_once!`, `at_most_once!` or have no acknowledgements at all without worrying about whether the other subscribers are going to re-publish the message.

Is there a way for us to setup the `at_least_once!` behavior so that it re-queues the message just to one subscriber's queue rather than re-publishing it to the exchange (and giving a copy to all of the subscribers listening to that routing key)?

mmmries:
This fixes #42 

Currently when a message goes into a queue with the `at_least_once!` behavior and fails, it gets re-published to the original exchange with the original routing key. This means the message will be re-queued for everyone listening to that routing key.

Since ActionSubscriber is intended to be focused on the subscriber and the mental model is of a single queue of work to be done, this behavior is very surprising.

This PR changes that behavior in a zero-downtime and backwards compatible way (but it will require some manual cleanup for people using the existing behavior). Rather than publishing to the original queue and with the original routing key, we setup a retry queue (with a slightly different naming scheme than the old retry queues) and publish a message into that retry queue (with a direct publishing using the default exchange). When that retry queue expires the message it will re-publish the message to the default exchange and use the original queue name as the routing key which will force it to be re-delivered only to that one queue.

A few things to consider:
* We can't change the behavior of the existing retry queues (set the `x-dead-letter-routing-key`), so we are generating new queues. After the old queues have drained they should be manually deleted from rabbitmq to avoid confusion.
* The new naming scheme uses a `.` for the separator which is much more consistent with how we build other queue names in ActionSubscriber
* When a message is re-delivered its `env` object will have a `routing_key = queue_name` rather than whatever the routing key was when it was originally published. This shouldn't matter for the vast majority of use cases, but it is possible that some people are using the `env.routing_key` to change some behavior.
  * This specifically matters to us because we are using the `routing_key` to report some metrics in `harness-action_subscriber` and I will need to push a gem update which uses the `queue` rather than the `routing_key` so that retries get counted the same as original messages.

/cc @quixoten @abrandoned @liveh2o @brianstien @localshred

### DIFF
--- a/lib/action_subscriber/bunny/subscriber.rb
+++ b/lib/action_subscriber/bunny/subscriber.rb
@@ -66,7 +66,7 @@ module ActionSubscriber
       def enqueue_env(threadpool, env)
         logger.info "RECEIVED #{env.message_id} from #{env.queue}"
         threadpool.async(env) do |env|
-          ::ActiveSupport::Notifications.instrument "process_event.action_subscriber", :subscriber => env.subscriber.to_s, :routing_key => env.routing_key do
+          ::ActiveSupport::Notifications.instrument "process_event.action_subscriber", :subscriber => env.subscriber.to_s, :routing_key => env.routing_key, :queue => env.queue do
             ::ActionSubscriber.config.middleware.call(env)
           end
         end
--- a/lib/action_subscriber/march_hare/subscriber.rb
+++ b/lib/action_subscriber/march_hare/subscriber.rb
@@ -67,7 +67,7 @@ module ActionSubscriber
       def enqueue_env(threadpool, env)
         logger.info "RECEIVED #{env.message_id} from #{env.queue}"
         threadpool.async(env) do |env|
-          ::ActiveSupport::Notifications.instrument "process_event.action_subscriber", :subscriber => env.subscriber.to_s, :routing_key => env.routing_key do
+          ::ActiveSupport::Notifications.instrument "process_event.action_subscriber", :subscriber => env.subscriber.to_s, :routing_key => env.routing_key, :queue => env.queue do
             ::ActionSubscriber.config.middleware.call(env)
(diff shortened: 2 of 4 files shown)


mmmries:
Should we release this change as a `2.0.0` or a `1.8.0`? It is backwards compatible and leaving the old retry queues hanging around won't break anything, but it would be really confusing to login to your Rabbitmq interface and see duplicate retry queues.

I think we should release it as a `2.0.0` and include a note about manually deleting the old retry queues in the changelog (which doesn't currently exist).

liveh2o:
This change seems reasonable to me, but I can't really comment on the specifics of the implementation. I'm fine with making this a major version bump, but manual queue cleanup is already a thing since queues are automatically created. I could really see this going either way.

film42:
:+1: This is a good catch. It might be a good idea to expose this as a configuration option in the future so anyone can pick the queue they want to retry to should something blow up.

`config.retry_with_queue_name { |ctx| ctx.full_queue_name }`

For now, I'm happy with a major version bump and with what I see. :shipit:

mmmries:
This has now been live in an MX production system for multiple days with no performance or other negative impact. I think we should go ahead and release this as `2.0.0`. Anyone want to veto?

vintagepenguin:
@mmmries looks like there are no vetos :shipit:

film42:
:shipit:

Issue #13, opened 2015-09-10
Configuration options should allow to be configured via connection string


brettallred:
Currently the only supported method of configuring Bunny or MarchHare is by a configuration parameters and not by a connection string.

mmmries:
Are you thinking we parse the connection string with the `URI` module and then set the corresponding settings in the configuration parameters? That seems like a pretty simple addition.

brettallred:
```ruby
      def connection_string
        AMQ::Settings.parse_amqp_url(::ActionSubscriber.configuration.connection_string)
      end
```

liveh2o:
I'm for it!

![ezgif-3518449088](https://cloud.githubusercontent.com/assets/207933/9806811/0cf4b8dc-5807-11e5-9634-735f9064f5b4.gif)

mmmries:
We don't have `AMQ` available to us, in this project. That is the eventmachine based implementation isn't it?

brettallred:
Bunny has a dependency on amq-protocol-2.0.0.   This is found in that gem.

bunny (2.2.0)
    amq-protocol (>= 2.0.0)

mmmries:
Nice. Sorry for the confusion.

Instead of checking for the `connection_string` when we are about to use the configuration, what if we just exploded the connection string and set the corresponding settings during the configuration step?

```ruby
module ActionSubscriber
  class Configuration
    ...
    def connection_string=(url)
      settings = ::AMQ::Settings.parse_amqp_url(url)
      settings.each do |key, value|
        send("#{key}=", value)
      end
    end
    ...
  end
end
```

brettallred:
:+1: I like it.

liveh2o:
Closed by #45.

PR #45, opened 2016-01-13
Adds support for connection string - Issue #13

brettallred:
See Issue #13

### DIFF
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -8,16 +8,16 @@ module ActionSubscriber
                   :host,
                   :hosts,
                   :mode,
-                  :pass,
+                  :password,
                   :pop_interval,
                   :port,
                   :prefetch,
                   :publisher_confirms,
-                  :user,
+                  :username,
                   :threadpool_size,
                   :timeout,
                   :times_to_pop,
-                  :vhost
+                  :virtual_host
 
     DEFAULTS = {
       :allow_low_priority_methods => false,
@@ -33,9 +33,9 @@ module ActionSubscriber
       :threadpool_size => 8,
       :timeout => 1,
       :times_to_pop => 8,
-      :user      => "guest",
-      :pass      => "guest",
-      :vhost     => "/"
+      :username => "guest",
+      :password => "guest",
+      :virtual_host => "/"
     }
 
     ##
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -32,7 +32,6 @@ module ActionSubscriber
       :port => 5672,
       :prefetch => 200,
       :publisher_confirms => false,
-      :ssl       => false,
       :threadpool_size => 8,
       :timeout => 1,
       :times_to_pop => 8,
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -8,13 +8,18 @@ module ActionSubscriber
                   :host,
                   :hosts,
                   :mode,
+                  :pass,
                   :pop_interval,
                   :port,
                   :prefetch,
                   :publisher_confirms,
+                  :user,
+                  :scheme,
+                  :ssl,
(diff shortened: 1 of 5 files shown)


lib/action_subscriber/configuration.rb
```diff
@@ -8,13 +8,18 @@ class Configuration
                   :host,
                   :hosts,
                   :mode,
+                  :pass,
```
film42: Password?

lib/action_subscriber/configuration.rb
```diff
@@ -8,13 +8,18 @@ class Configuration
...
                   :prefetch,
                   :publisher_confirms,
+                  :user,
+                  :scheme,
+                  :ssl,
                   :threadpool_size,
                   :timeout,
-                  :times_to_pop
+                  :times_to_pop,
+                  :vhost
```
film42: `virtual_host`?

mmmries: :+1: for expanding to the full names here

lib/action_subscriber/configuration.rb
```diff
@@ -79,6 +88,13 @@ def add_decoder(decoders)
       self.decoder.merge!(decoders)
     end
 
+    def connection_string=(url)
+      settings = ::AMQ::Settings.parse_amqp_url(url)
+      settings.each do |key, value|
+        send("#{key}=", value)
```
mmmries: if we use the full names then this probably won't work huh? Because `::AMQ::Settings` will try to pass the `key` as `vhost`, `pass`, etc...

brettallred: Correct.

brettallred: Which direction do you want to go?

mmmries:
I'm okay to use the short names, have you checked that the short names are accepted by both `march_hare` and `bunny`? Connection options are one of the places where I've seen some differences between the two libraries.

Also would you mind adding a test example of how you turn on ssl via the connection string?

brettallred:
I validated they are accepted by both

For reference:
http://rubymarchhare.info/articles/connecting.html
http://rubybunny.info/articles/connecting.html

brettallred:
Good call on the SSL.  That doesn't work as a connection option. I  removed it as a connection option since we don't need ssl support and  it doesn't look like you guys to either. If we want to support SSL there are other things we will need to add to the gem so I figure its best to leave it out for now.

film42:
:shipit:

lib/action_subscriber/configuration.rb
```diff
@@ -8,13 +8,18 @@ class Configuration
...
                   :hosts,
                   :mode,
+                  :pass,
                   :pop_interval,
                   :port,
                   :prefetch,
                   :publisher_confirms,
+                  :user,
+                  :scheme,
+                  :ssl,
```
film42: Let's get rid off this too, then.

brettallred: If we get rid of that then we can't explode the parse response into settings like @mmmries wanted to.
```
settings = ::AMQ::Settings.parse_amqp_url(url)
settings.each do |key, value|
       send("#{key}=", value)
end
```

mmmries: Hmmmm, it seems very weird to have a config option for `ssl`, but not actually support it. Can we do something like `settings.slice(:vhost, :user, :pass, ...)` so we just ignore the `ssl` part of it?

brettallred:
@mmmries - How about this?  I just brought the parser into Action Subscriber and modified it.  Now all the naming is per the suggestions, etc.

mmmries:
:+1: looks good to me. Glad that the URI parser was pretty small

:shipit:

mmmries:
BTW we have a pre-release version of the gem where we are testing out some asynchronous publishing, once that is done testing (it is in our production system right now) we will cut a new version that will include this change. Thanks for the PR @brettallred

mmmries:
#PresidentsCanCodeToo

PR #46, opened 2016-01-25
Better default values

mmmries:
In our production environment we've switched the default `prefetch` to 5 across most of our subscribers and it has worked well for us. Prefetching 200 messages (per queue) means you end up holding onto a lot of messages and they can end up sitting around for a while before getting used.

If your message processing is in the 10ms+ range then prefetching that many messages does not actually speed things up so I've bumped the default down for the gem.

I've also setup a configuration value for how many seconds we should try to wait for the threadpool to drain during graceful shutdowns and defaulted it to 30 seconds (instead of the previous hardcoded 1000 seconds).

/cc @abrandoned @localshred @brettallred @brianstien @quixoten

### DIFF
--- a/lib/action_subscriber/babou.rb
+++ b/lib/action_subscriber/babou.rb
@@ -96,8 +96,7 @@ module ActionSubscriber
       wait_loops = 0
       ::ActionSubscriber::Babou.stop_receving_messages!
 
-      # Going to wait until the thread pool drains or we wait for 1000 seconds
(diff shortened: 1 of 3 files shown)


quixoten:
That's a long config parameter. Would  `shutdown_timeout` or `graceful_shutdown_timeout` offer the same clarity? Either way, you've got my :shipit:

mmmries:
@quixoten it is pretty long, I just wanted to make it obvious that it was measured in seconds (not milliseconds like some other config parameters)

liveh2o:
:shipit:

PR #47, opened 2016-01-26
Fix Action Bug

mmmries:
There is an existing bug in ActionSubscriber where the user can setup a route with `action = :some_method`, but when messages are received it will use the last part of the `routing_key` instead of the action that was specific when the user was drawing their routes.

I'm almost positive this was an oversight when we built in the explicit routing layer because it is extremely surprising when you setup a route like `route ::AggregationSubscriber, :completed, ...`, but in production it tries to call `completed_low` because the routing key was specific as `grunt.aggregation.completed_low`.

/cc @abrandoned @brettallred @localshred @film42 @quixoten


liveh2o:
Seems fine to me. :shipit:

PR #48, opened 2016-02-03
Add a durable asynchronous publisher

film42:
**What:**
This is a work-in-progress PR that allows you to pull message publishing into its own supervised thread. It's important to note that I'm not trying to change the `publish` code, instead I'm proposing to add a new layer of abstraction on top of it to add durability to message publishing. This PR will be the first of a few:

1. Establish a pattern for asynchronous message publishing (this PR).
2. Create a pattern that persists messages to disk (using redis).
3. Extract publisher code to its own gem.

With these changes, you will be able to durably and reliably publish messages without needing to rely on an immediately available RabbitMQ instance. For example, if you're publishing changes in your application's request path, a sick RabbitMQ cluster won't pose an immediate threat.

Writing these messages to disk is the main goal I'm trying to get to. I'm starting with an in-memory buffer because it's a simple default that we can ship in ActionSubscriber without needing to change our requirements. It also means that adding a redis layer becomes a simple configuration change. If you're not feeling too crazy about an in-memory buffer and think we should skip ahead to the redis version, please share your feedback.

Note: This doesn't include tests at this point, but those are coming. I want to make sure this looks like a good path forward before putting those up.

**Performance:**
Locally I can push ~500 1KB messages/sec on MRI, and ~900 1KB messages/sec on Jruby. I think these numbers are pretty good for a single-threaded publisher. If this becomes a problem, we can always create a pool of publishers.

cc @mmmries @abrandoned @liveh2o @brianstien @localshred

### DIFF
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -23,6 +23,7 @@ require "action_subscriber/bunny/subscriber"
 require "action_subscriber/march_hare/subscriber"
 require "action_subscriber/babou"
 require "action_subscriber/publisher"
+require "action_subscriber/publisher/async"
 require "action_subscriber/route"
 require "action_subscriber/route_set"
 require "action_subscriber/router"
@@ -109,6 +110,9 @@ module ActionSubscriber
 
   # Initialize config object
   config
+
+  # Intialize async publisher adapter
+  ::ActionSubscriber::Publisher::Async.publisher_adapter
 
   ::ActiveSupport.run_load_hooks(:action_subscriber, Base)
 
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -139,5 +139,6 @@ end
 require "action_subscriber/railtie" if defined?(Rails)
 
 at_exit do
+  ::ActionSubscriber::Publisher::Async.publisher_adapter.shutdown!
   ::ActionSubscriber::RabbitConnection.publisher_disconnect!
 end
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -2,6 +2,7 @@ module ActionSubscriber
   class Configuration
     attr_accessor :allow_low_priority_methods,
                   :async_message_publisher,
+                  :async_message_supervisor_interval,
                   :decoder,
                   :default_exchange,
                   :error_handler,
@@ -21,6 +22,7 @@ module ActionSubscriber
     DEFAULTS = {
       :allow_low_priority_methods => false,
       :async_message_publisher => 'memory',
+      :async_message_supervisor_interval => 200, # in milliseconds
       :default_exchange => 'events',
       :heartbeat => 5,
       :host => 'localhost',
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -1,8 +1,10 @@
 module ActionSubscriber
   class Configuration
     attr_accessor :allow_low_priority_methods,
-                  :async_message_publisher,
-                  :async_message_supervisor_interval,
+                  :async_publisher,
+                  :async_publisher_drop_messages_when_queue_full,
+                  :async_publisher_max_queue_size,
+                  :async_publisher_supervisor_interval,
                   :decoder,
                   :default_exchange,
                   :error_handler,
@@ -21,8 +23,10 @@ module ActionSubscriber
 
     DEFAULTS = {
       :allow_low_priority_methods => false,
-      :async_message_publisher => 'memory',
-      :async_message_supervisor_interval => 200, # in milliseconds
+      :async_publisher => 'memory',
+      :async_publisher_drop_messages_when_queue_full => false,
+      :async_publisher_max_queue_size => 1_000_000,
+      :async_publisher_supervisor_interval => 200, # in milliseconds
       :default_exchange => 'events',
       :heartbeat => 5,
       :host => 'localhost',
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -1,6 +1,7 @@
 module ActionSubscriber
   class Configuration
     attr_accessor :allow_low_priority_methods,
+                  :async_message_publisher_mode,
                   :decoder,
                   :default_exchange,
                   :error_handler,
@@ -19,6 +20,7 @@ module ActionSubscriber
 
     DEFAULTS = {
       :allow_low_priority_methods => false,
+      :async_message_publisher_mode => 'memory',
       :default_exchange => 'events',
       :heartbeat => 5,
       :host => 'localhost',
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -1,7 +1,7 @@
 module ActionSubscriber
   class Configuration
     attr_accessor :allow_low_priority_methods,
-                  :async_message_publisher_mode,
+                  :async_message_publisher,
                   :decoder,
                   :default_exchange,
                   :error_handler,
@@ -20,7 +20,7 @@ module ActionSubscriber
 
     DEFAULTS = {
       :allow_low_priority_methods => false,
-      :async_message_publisher_mode => 'memory',
+      :async_message_publisher => 'memory',
       :default_exchange => 'events',
       :heartbeat => 5,
       :host => 'localhost',
--- a/lib/action_subscriber/publisher/async.rb
+++ b/lib/action_subscriber/publisher/async.rb
@@ -1,0 +1,31 @@
+module ActionSubscriber
+  module Publisher
+    # Publish a message asynchronously to RabbitMQ.
+    #
+    # Asynchronous is designed to do two things:
+    # 1. Introduce the idea of a durable retry should the RabbitMQ connection disconnect.
+    # 2. Provide a higher-level pattern for fire-and-forget publishing.
+    #
+    # @param [String] route The routing key to use for this message.
+    # @param [String] payload The message you are sending. Should already be encoded as a string.
+    # @param [String] exchange The exchange you want to publish to.
+    # @param [Hash] options hash to set message parameters (e.g. headers).
+    def self.publish_async(route, payload, exchange_name, options = {})
+      Async.publisher_adapter.publish(route, payload, exchange_name, options)
+    end
+
+    module Async
+      def self.publisher_adapter
+        @publisher_adapter ||= case ::ActionSubscriber.configuration.async_message_publisher_mode
+                               when /memory/i then
+                                 require "action_subscriber/publisher/async/in_memory_adapter"
+                                 InMemoryAdapter.new
+                               when /redis/i then
+                                 fail "Not yet implemented"
+                               else
+                                 fail "Unknown async publisher provided"
+                               end
+      end
+    end
+  end
+end
--- a/lib/action_subscriber/publisher/async.rb
+++ b/lib/action_subscriber/publisher/async.rb
@@ -16,7 +16,7 @@ module ActionSubscriber
 
     module Async
       def self.publisher_adapter
-        @publisher_adapter ||= case ::ActionSubscriber.configuration.async_message_publisher
+        @publisher_adapter ||= case ::ActionSubscriber.configuration.async_publisher
                                when /memory/i then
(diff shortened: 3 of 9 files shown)


lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+          fail UnableToPersistMessageError, "Queue is full, messages will be dropped." if queue.size > MAXIMUM_QUEUE_SIZE
+
+          queue.push(message)
+        end
+
+      private
+
+        def create_and_supervise_consumer!
+          @consumer = create_consumer
+          @supervisor = ::Thread.new do
```
abrandoned: how do these threads die and get collected? supervisor just run forever?

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+              @current_message = message = queue.pop
+
+              begin
+                ::ActionSubscriber::Publisher.publish(message.route, message.payload, message.exchange_name, message.options)
+
+                # Reset
+                @current_message = nil
+              rescue *NETWORK_ERRORS
+                # Sleep because the connection is down.
+                sleep 1
```
abrandoned: allow configurable sleep millis? (as opposed to seconds)

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+                # Requeue and try again.
+                queue.push(message)
+              rescue => unknown_error
+                # Do not requeue the message because something else horrible happened.
+                @current_message = nil
+                puts unknown_error.class
```
abrandoned: don't we have a logger in action_subscriber?

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+            loop do
+              unless consumer.alive?
+                # Why might need to requeue the last message.
+                queue.push(@current_message) if @current_message.present?
+                consumer.kill
+                @consumer = create_consumer
+              end
+
+              # Pause before checking the consumer again.
+              sleep 1
```
abrandoned: same here with configurable millis?

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+
+        def initialize
+          @queue = ::Queue.new
+          create_and_supervise_consumer!
+        end
+
+        def push(message)
+          # TODO: How do we exert backpressure here? Do we block the thread, or do we raise an error?
+          fail UnableToPersistMessageError, "Queue is full, messages will be dropped." if queue.size > MAXIMUM_QUEUE_SIZE
```
abrandoned: maybe a configurable `fail` vs just throwing away?

abrandoned:
would be interesting to see a test with lifeguard (even set at 5 threads)

lib/action_subscriber/configuration.rb
```diff
@@ -1,6 +1,7 @@
 module ActionSubscriber
   class Configuration
     attr_accessor :allow_low_priority_methods,
+                  :async_message_publisher_mode,
```
liveh2o: Let's call this `async_message_publisher`.

lib/action_subscriber/publisher/async.rb
```diff
@@ -0,0 +1,31 @@
...
+    # @param [String] exchange The exchange you want to publish to.
+    # @param [Hash] options hash to set message parameters (e.g. headers).
+    def self.publish_async(route, payload, exchange_name, options = {})
+      Async.publisher_adapter.publish(route, payload, exchange_name, options)
+    end
+
+    module Async
+      def self.publisher_adapter
+        @publisher_adapter ||= case ::ActionSubscriber.configuration.async_message_publisher_mode
+                               when /memory/i then
```
liveh2o: Why use a regex here? It seems reasonable that a lowercase string would need to be used in the config. Alternatively, you could require a symbol.

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+      class UnableToPersistMessageError < ::StandardError
+      end
+
+      class AsyncQueue
+        attr_reader :consumer, :queue, :supervisor
+
+        MAXIMUM_QUEUE_SIZE = 1_000_000.freeze
+
+        if ::RUBY_PLATFORM == "java"
+          NETWORK_ERRORS = [::Java::ComRabbitmqClient::AlreadyClosedException, ::Java::JavaIo::IOException]
```
liveh2o: Perhaps freeze these as well/

lib/action_subscriber/publisher/async/in_memory_adapter.rb
film42: I uses 1 second, because that's what we set our `network_recovery_interval` to [here](https://github.com/mxenabled/action_subscriber/blob/f818f77f9a9fe85aaf27905623236158db1ca363/lib/action_subscriber/rabbit_connection.rb#L69). Though not currently configurable, perhaps it should be.

lib/action_subscriber/publisher/async/in_memory_adapter.rb
film42: Supervisor and consumer are collected when the program terminates (run forever). Maybe a shutdown trigger is needed so it can empty its queue before terminating? Actually, there probably is a possible race condition where the request finishes and exits the process, but the publisher has not successfully delivered the message. I'm going to need to think about this.

liveh2o:
Overall, I'm really liking the direction. I think starting with an in-memory queue is a solid default that will allow anyone to get this up and running with little overhead. The adapter interface seems pretty solid as well.

lib/action_subscriber/publisher/async/in_memory_adapter.rb
film42: That's a good idea. Basically saying, "you can either have an error raised because you've saturated your publisher, or you can drop your messages."

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,153 @@
...
+          end
+
+          def size
+            queue.size
+          end
+
+        private
+
+          def await_network_reconnect
+            sleep ::ActionSubscriber::RabbitConnection::NETWORK_RECOVERY_INTERVAL
```
film42: @abrandoned I'm using this constant to be the first path towards configuring network recover intervals.

lib/action_subscriber.rb
```diff
@@ -135,5 +139,6 @@ def self.default_routes
 require "action_subscriber/railtie" if defined?(Rails)
 
 at_exit do
+  ::ActionSubscriber::Publisher::Async.publisher_adapter.shutdown!
```
film42: I'm not sure that I love using `at_exit` to drain the pool, but it will terminate after `seconds_to_wait_for_graceful_shutdown` has expired with nice messages.

film42:
I removed the RFC from this PR and added tests, so this is ready for a proper review.

mmmries:
What are the pros/cons of putting this work into `action_subscriber` vs starting the `active_publisher` gem we've been talking about? It seems easier to me to release these changes as `active-publisher v0.0.1` rather than release a pre-release version of ActionSubscriber.

lib/action_subscriber/publisher/async/in_memory_adapter.rb
mmmries: I'm okay leaving this as a constant because if we end up using any async-adapter that goes to redis we probably won't be in charge of configuring the network recovery intervals inside the publishing process.

mmmries:
@abrandoned would moving to a threadpool of publishing threads be useful in this project? If we end up using redis + sidekiq we will lose control of the threadpool execution anyway. I think the only real benefit we gain from a threadpool is that the workers in the threadpool wouldn't have to open/close a channel on each publish which should reduce the overhead for the publisher and the rabbit cluster, but that definitely sounds like a separate iteration to me.

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -0,0 +1,153 @@
...
+                rescue => unknown_error
+                  # Do not requeue the message because something else horrible happened.
+                  @current_message = nil
+
+                  # Log the error.
+                  logger.info unknown_error.class
+                  logger.info unknown_error.message
+                  logger.info unknown_error.backtrace.join("\n")
+
+                  # TODO: Find a way to bubble this out of the thread for logging purposes.
```
jtanner: One approach could be to use something like [Celluloid Actor Linking](https://github.com/celluloid/celluloid/wiki/Linking) and `trap_exit` to catch and notify the supervisor actor of the error.

abrandoned:
@mmmries all for iteration; no need to pool it yet (until we have some kind of data about what would be good)

lib/action_subscriber/publisher/async/in_memory_adapter.rb
film42: Yeah, I think we need to provide a way to jack-in an error handler here. I'm willing to punt on this for now, though.

abrandoned:
I'm all for taking this for a spin and seeing what we come up with :shipit:

film42:
Going to cut a pre-release.

mmmries:
Not necessary for this PR, but I was just thinking that when we use `publish_async` it would probably be a good idea to default to publishing with `publisher_confirms` turned on, so that we can retry publishes that fail to reach the cluster

film42:
I like that idea, @mmmries .

abrandoned:
@film42 is this ready to go? let's make it go if so :shipit:

film42:
@abrandoned This is ready to go. I'll follow up this PR with my introspection strategy which will allow us to publish these inner stats using `harnes-action_subscriber`.

spec/lib/action_subscriber/configuration_spec.rb
```diff
@@ -1,7 +1,10 @@
 describe ::ActionSubscriber::Configuration do
   describe "default values" do
     specify { expect(subject.allow_low_priority_methods).to eq(false) }
-    specify { expect(subject.default_exchange).to eq("events") }
```
mmmries: why remove the test for `default_exchange`?

spec/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
 require 'action_subscriber/rspec'
 
 # Silence the Logger
+$TESTING = true
```
mmmries: Did you end up using this anywhere? I don't see it anywhere else in the diff.

mmmries:
A few small questions of things that I think we could cleanup, but :shipit: regardless of the response. Neither of them are worth holding off for.

spec/lib/action_subscriber/configuration_spec.rb
film42: Whoops! That should not be removed!

film42: Done.

spec/spec_helper.rb
film42: I had to set this so we wouldn't attempt to close the `oldlogger` in Logging.

film42:
Btw, for those who are interested, here's a snippet you can use to introspect the running async publisher:
```ruby
trap(:TRAP) do
  queue_size = ::ActionSubscriber::Publisher::Async.publisher_adapter.async_queue.size
  supervisor = ::ActionSubscriber::Publisher::Async.publisher_adapter.async_queue.supervisor
  consumer = ::ActionSubscriber::Publisher::Async.publisher_adapter.async_queue.consumer
  ::Rails.logger.info "ASYNC PUBLISHER QUEUE SIZE: #{queue_size}"
  ::Rails.logger.info "  SUPERVISOR: #{supervisor.alive? ? 'ALIVE' : 'DEAD' }"
  ::Rails.logger.info "  CONSUMER: #{consumer.alive? ? 'ALIVE' : 'DEAD' }"
end
```

A better solution for introspecting is coming soon.

PR #49, opened 2016-02-10
Provide error handler to the async publisher

film42:
Adding an error handler to the async publisher will let anyone bubble errors wherever they need to go. By default, however, exceptions are logged and swallowed (not new behavior).

I think this PR is in an adequate state. What I'm unsure of is if this behavior should become a convention of action subscriber (set the default in the configuration), or leave this in the in-memory publisher so it's easier to extract into a publisher gem. You might think it's good enough as is, and that's cool too.

### DIFF
--- a/lib/action_subscriber/publisher/async/in_memory_adapter.rb
+++ b/lib/action_subscriber/publisher/async/in_memory_adapter.rb
@@ -65,6 +65,20 @@ module ActionSubscriber
             create_and_supervise_consumer!
           end
 
+          def error_handler
+            # This is currently not memoized because the configuration might change by some other
+            # gem loaded after action subscriber starts publishing. Because this code path only
+            # executes when something terrible happens, I think this trade-off is fine.
+            handler = ::ActionSubscriber.configuration.async_publisher_error_handler
+
+            if !handler.respond_to?(:call) || handler.arity != 1
+              logger.error "Error handler must respond to #call with an arity of 1"
+              return
+            end
+
+            handler
+          end
+
           def push(message)
             # Default of 1_000_000 messages.
             if queue.size > ::ActionSubscriber.configuration.async_publisher_max_queue_size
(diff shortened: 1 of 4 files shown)


film42:
P.S. Not sure what's up with the failing spec (seems unrelated). cc @mmmries

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -65,6 +65,20 @@ def initialize
             create_and_supervise_consumer!
           end
 
+          def error_handler
+            # This is currently not memoized because the configuration might change by some other
+            # gem loaded after action subscriber starts publishing. Because this code path only
+            # executes when something terrible happens, I think this trade-off is fine.
+            handler = ::ActionSubscriber.configuration.async_publisher_error_handler
+
+            if !handler.respond_to?(:call) || handler.arity != 1
```
mmmries: I'm wondering if we should check for the `respond_to?` and `arity` when we are setting it in the `Configuration` object rather than checking it on each exception. What do you think?

film42: I agree 100%. I'm just not sure if delegating that to ActionSubscriber is going to make this harder to extract.

liveh2o: It shouldn't be any harder to extract as long as it's scope is limited to the configuration object. Let's go ahead and do it now.

liveh2o:
One change, and this should be G2G.

film42:
Should be good to go.

liveh2o:
:shipit:

mmmries:
I think we should probably make this error handler accept 2 arguments instead of 1 [the same way we do for our current error handler](https://github.com/mxenabled/action_subscriber/blob/58001409b86588fdbd7d257b5183942d8a99fe60/lib/action_subscriber/middleware/error_handler.rb#L13).

This allows you to pass some contextual information (like the message we were trying to publish perhaps) which is really helpful to understand the backtrace.

film42:
@mmmries I think that's fine. My next question, then, is should we have two error handlers, or just one?

brettallred:
It seems like there should be just one error handler. I don't see the use case in which you would want to configure two different error handlers, one for async publishing and on for synchronous publishing. Unless... having two will make it easier to extract the publisher gem, in which case my vote would be to keep the error handler api the same in the sense that it accepts takes 2 params, not just 1.

mmmries:
Similar to #50 I think we can close this now that `active_publisher` has been spun off

PR #50, opened 2016-02-10
Add instrumentation to async publisher for tracking metrics

film42:
With this change, we can now subscribe to supervisor ticks as they happen in the async publisher. It will send you its queue size and whether or not the consumer is alive. You could, for example, use this to log information about the publisher to statsd.

A future step would be to connect the work-in-progress error handler to provide even more insight into what's happening inside action subscriber (or specifically what happens when someone subscribing blows up).

### DIFF
--- a/lib/action_subscriber/publisher/async/in_memory_adapter.rb
+++ b/lib/action_subscriber/publisher/async/in_memory_adapter.rb
@@ -52,7 +52,7 @@ module ActionSubscriber
         class AsyncQueue
           include ::ActionSubscriber::Logging
 
-          attr_reader :consumer, :queue, :supervisor
+          attr_reader :consumer, :observer, :queue, :supervisor
 
           if ::RUBY_PLATFORM == "java"
             NETWORK_ERRORS = [::MarchHare::Exception, ::Java::ComRabbitmqClient::AlreadyClosedException, ::Java::JavaIo::IOException].freeze
@@ -62,7 +62,7 @@ module ActionSubscriber
 
           def initialize
             @queue = ::Queue.new
-            create_and_supervise_consumer!
+            create_and_supervise_consumer_observer
           end
 
           def push(message)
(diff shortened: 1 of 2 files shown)


lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -62,6 +62,7 @@ class AsyncQueue
 
           def initialize
             @queue = ::Queue.new
+            @instrumentation_threadpool = ::Lifeguard::Threadpool.new
```
film42: I could make this a single thread because Lifeguard won't block, but I left this as default for now.

mmmries: I would prefer to see the `::ActiveSupport::Notifications.instrument` get called inline rather than put into a threadpool. We will end up using harness which is not really a concern for blocking the main thread (or the supervisor thread in this case).

If we don't need the threadpool then I would vote to remove it.

film42: @mmmries I used a threadpool because I don't like the idea of external code degrading the reliability of the supervisor, whether it be slowing the interval down or an exception coming back up through the instrument. I originally wrote this without one, but added this because of my previously stated concerns. I'm happy to rip it out if you think it's safe to do so.

localshred:
FWIW I wrote a statsd/railsy plugin thingy that uses ActiveSupport notifications under the hood. It has a sync and async publisher (using concurrent-ruby). May be useful.

https://github.com/localshred/nsa

film42:
@localshred That looks super cool! We've been using [`harness-action_subscriber`](https://github.com/mxenabled/harness-action_subscriber) so far.

localshred:
Va Va Voom :car: :dash:

lib/action_subscriber/publisher/async/in_memory_adapter.rb
```diff
@@ -92,11 +93,20 @@ def create_and_supervise_consumer!
...
+                    queue.push(@current_message) if @current_message.present?
+                    consumer.kill
+                    @consumer = create_consumer
+                  end
+
+                  instrumentation_threadpool.async do
+                    payload = { :queue_size => size, :is_consumer_alive => consumer.alive? }
+                    ::ActiveSupport::Notifications.instrument "supervisor_tick.in_memory_publisher.action_subscriber", payload
+                  end
+                rescue
```
mmmries: the unqualified `rescue` makes me a bit nervous here. Are we guarding against having `ActiveSupport::Notifications.instrument` raise an exception?

film42: @mmmries I just want to ensure that the supervisor never dies. So this ensures things keep ticking. I agree it's a bit weird, though.

mmmries:
Here's an alternate implementation that you might have already discarded, but what do you think of doing something like:

```ruby
class AsyncQueue
 def initialize
    @observer = create_observer
  end

  def create_observer
    async_queue = self
    ::Thread.new do
      loop do
        ::ActiveSuportBlahBlah.instrument(...)
        sleep 0.5
      end
    end
  end
end
```

I'm imagining having a thread that sits around and measures the queue every so often, but it doesn't actually run in the supervisor thread. That way we don't have to worry about it degrading the performance and if it crashes unexpectedly we just lost the instrumentation, we don't lose any functionality.

film42:
@mmmries I like that a lot better. I can also make the supervisor check on the observer so it can recover.

mmmries:
We ended up moving all the publishing code to `active_publisher` so I think we can close this now?

Issue #51, opened 2016-02-19
draw_routes is ignoring config


film42:
When drawing explicit routes, CLI and yaml configs are ignored.

**Example:**
```ruby
ENV["APP_NAME"] = "yolo"

require "action_subscriber"

class UserSubscriber < ::ActionSubscriber::Base
  def created
  end

  def created_low
  end
end

::ActionSubscriber.draw_routes do
  default_routes_for ::UserSubscriber
end

::Thread.new do
  # Wait for subscribers to startup and then check the thread pool
  sleep 2
  puts ::ActionSubscriber::Threadpool.pool(:default).inspect
end
```

```
$ action_subscriber start --allow-low-priority-methods --threadpool-size 40 --prefetch 5
```

```
I, [2016-02-18T21:19:13.345970 #49177]  INFO -- : Starting in-memory publisher adapter.
I, [2016-02-18T21:19:13.346747 #49177]  INFO -- : Loading configuration...
I, [2016-02-18T21:19:13.346901 #49177]  INFO -- : Starting server...
I, [2016-02-18T21:19:13.356552 #49177]  INFO -- : Rabbit Host: localhost
Rabbit Port: 5672
Threadpool Size: 40
Low Priority Subscriber: true
Decoders:
  --application/json
  --text/plain
I, [2016-02-18T22:35:06.379808 #61299]  INFO -- : UserSubscriber
I, [2016-02-18T22:35:06.379836 #61299]  INFO -- :   -- method: created
I, [2016-02-18T22:35:06.379856 #61299]  INFO -- :     --    exchange: events
I, [2016-02-18T22:35:06.379875 #61299]  INFO -- :     --       queue: yolo.user.created
I, [2016-02-18T22:35:06.379892 #61299]  INFO -- :     -- routing_key: user.created
I, [2016-02-18T22:35:06.379914 #61299]  INFO -- : Action Subscriber connected

#<Lifeguard::InfiniteThreadpool:0x007f9d7a09f4d8 @pool_size=8>
```

So despite having a `created_low` queue, and setting the thread pool to 40, both options are ignored.

liveh2o:
Was this resolved by #53, @film42?

PR #52, opened 2016-02-19
Require app after configuration has been loaded

film42:
Because of how we're setting up routes to be drawn (on app load), we're
actually building the routes using a configuration that is using a
default config. However, when we infer routes, we're delaying this route
building to after the configuration has been loaded from CLI and yaml.

This means that #default_routes_for will not create *_low queues or set
real threadpool values despite your CLI or yaml config. If we require
the app after the ActionSubscriber gem has been configured, everything
works as expected.

Although this is a small code change and I don't imagine anyone having
problems, we are changing execution order, which is a pretty big design
change.

This is one of a few ways at fixing https://github.com/mxenabled/action_subscriber/issues/51

### DIFF
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -3,6 +3,7 @@
 require 'active_support'
 require 'active_support/core_ext'
 require 'thor'
+require 'action_subscriber'
 
 module ActionSubscriber
   class CLI < ::Thor
@@ -23,12 +24,14 @@ module ActionSubscriber
     BABOUDESC
 
     def start
-      require options[:app]
-
       $0 = "Action Subscriber server #{object_id}"
       ::ActionSubscriber.logger.info "Loading configuration..."
 
       ::ActionSubscriber::Configuration.configure_from_yaml_and_cli(options)
+
+      ::ActionSubscriber.logger.info "Requiring app..."
+      require options[:app]
+
       ::ActionSubscriber.logger.info "Starting server..."
 
       case ::ActionSubscriber.configuration.mode
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 
+$ACTION_SUBSCRIBER_SERVER_MODE = true
+
 require 'active_support'
 require 'active_support/core_ext'
 require 'thor'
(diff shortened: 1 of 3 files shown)


liveh2o:
@mmmries and I had a conversation about how the binstub expects the app to handle loading the rest of the gem a few months ago. We had an app that wasn't requiring the Action Subscriber gem, so starting the subscriber would fail. At the time, he mentioned that loading the app first resolved some issues we had in the gem early on, but he couldn't remember exactly what those problems were.

I'd say we have a compelling reason to make this change, but with that context, I'd say we should either push a pre-release that we can thoroughly test to see if there any unintended side affects, or we should explore one of the other ways of fixing the problem you mentioned.

Another option I'd suggest is possibly deferring drawing the routes until `auto_pop!` or `start_subscribers` are called. Since we capture all of the routes in a block, that should be possible, though I haven't looked at the code to determine if there are any other edge cases to consider.

film42:
I really want to push this forward, because it turns out that creating a threadpool outside of the `draw_routes` will load the config and setup the default threadpool too early (with the default size). This really confusing to debug, so loading things up ahead of time would be a major win for clarity.

liveh2o:
I'm fine with the change, if @mmmries is. What say ye, @mmmries?

mmmries:
I think I remember making this change once before and then hotfixing it, but I can't remember what went wrong, I'm going to check the history of PRs on this project really quick

mmmries:
Found the [thing I was thinking of](https://github.com/mxenabled/action_subscriber/pull/28) and it shouldn't affect this PR. It was a load order issue around requiring the `action_suscriber.rb` before the app had loaded so `defined? Rails` would alway be `false`.

I think that loading the ActionSubscriber configured (CLI and YAML) should be safe to do before loading the application.

:shipit:

film42:
Before I merge this I want to double check a few of our apps to make sure they start just fine with these changes.

film42:
Had to require `yaml` in the config and drop the `safe_yaml` stuff (but it's 2-3 years old).

mmmries:
:+1: for dropping the `:safe_yaml`. We won't have loaded all of the rails libraries at this point, but since it is a yaml file we control we don't need the extra safety check right now.

film42:
Ready to go!

mmmries:
:shipit:

film42:
@mmmries this will introduce the same problem, but we can add the railtie to the binstub to avoid this problem. We can probably remove it from `lib/action_subscriber.rb` (doesn't look like the publisher code needs them).

```
::ActionSubscriber.logger.info "Requiring app..."
require options[:app]
require "action_subscriber/railtie" if defined?(Rails)
```

mmmries:
@film42 nice catch. I think the conditional require in the CLI sounds like a good option. Let's just test it with a few of our apps to verify.

film42:
@mmmries I moved the railtie to the bin stub.

mmmries:
@film42 did a few tests locally with one of our apps and the empirical tests look good as well as the theory here.

:shipit:

film42:
I think we're going to need to make one more change befor this is finally ready.

film42:
Added a few more changes and tests both server-side loading scenarios and client-side loading scenarios. Everything is working :ok_hand: .

mmmries:
I just decided that getting ActivePublisher going is now a high priority for me :pensive: 

We talked offline about this, but obviously this is a case where we have gotten ourselves into a bit of a pickle and this is the best path forward to keep backwards compatibility for our clients. Once we split out the publishing side of this to its own gem we can get rid of the conditional require + load hooks

:shipit:

film42:
Thanks a bunch for your help @mmmries !

Released as `2.2.0`.

PR #53, opened 2016-02-19
Delay ActionSubscriber.draw_routes execution until after config loads

film42:
Because the app is loaded before ActionSubscriber configs, immediate
execution of the draw_routes block means routes will be built using the
default configs. This means threadpool settings and low priority queues
will not be included.

This is one of a few ways at fixing #51 

### DIFF
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -60,8 +60,12 @@ module ActionSubscriber
   end
 
   def self.draw_routes(&block)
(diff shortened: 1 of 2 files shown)


liveh2o:
Changes look good, but specs are failing of some reason.

film42:
Updated the inference spec to bust memoized routes before running.

liveh2o:
:shipit:

film42:
Released as 2.1.1

PR #54, opened 2016-02-20
Multi-threadpool bugs

film42:
This handles two different multi-threadpool bugs. They're small enough that I combined them together into one PR.

**1) Check any threadpool for room before skipping a round of pops (pop mode)**

In pop mode, we check to see if all pools have space before and won't check for new messages otherwise. This used to be fine but now that we allow multiple threadpools, we consider all threadpools full if any one of them is full, even though a larger pool may be empty.

To fix this, we flip the .busy? logic check to say, "we are .ready? for popping if any of the threadpools has room." We then need to add a guard clause around each queue when popping to skip if its threadpool is already full. This will ensure all threadpools are able to fill completely.

**2) Allow graceful shutdown check to work with multiple threadpools**

With this change we'll make sure all threadpools are empty before shutting down. It will only list threadpools that are not empty to avoid extra noise.

The new shutdown message looks like this:
```
waiting for threadpools to empty:
  -- default (remaining: 40)
  -- some_custom_pool (remaining: 2)
  -- yolo (remaining: 5)
threadpool empty. Shutting down
```

cc @mmmries @abrandoned @liveh2o @brianstien @localshred @quixoten @ztoolson @brettallred

### DIFF
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -81,6 +81,7 @@ module ActionSubscriber
         logger.info "    --    exchange: #{route.exchange}"
         logger.info "    --       queue: #{route.queue}"
         logger.info "    -- routing_key: #{route.routing_key}"
+        logger.info "    --  threadpool: #{route.threadpool.pool_size}"
       end
     end
   end
--- a/lib/action_subscriber.rb
+++ b/lib/action_subscriber.rb
@@ -81,7 +81,7 @@ module ActionSubscriber
         logger.info "    --    exchange: #{route.exchange}"
         logger.info "    --       queue: #{route.queue}"
         logger.info "    -- routing_key: #{route.routing_key}"
-        logger.info "    --  threadpool: #{route.threadpool.pool_size}"
+        logger.info "    --  threadpool: #{route.threadpool.name}, pool_size: #{route.threadpool.pool_size}"
       end
     end
   end
--- a/lib/action_subscriber/bunny/subscriber.rb
+++ b/lib/action_subscriber/bunny/subscriber.rb
@@ -17,6 +17,9 @@ module ActionSubscriber
         times_to_pop = [::ActionSubscriber::Threadpool.ready_size, ::ActionSubscriber.config.times_to_pop].min
         times_to_pop.times do
           queues.each do |route, queue|
+            # Handle busy checks on a per threadpool basis
+            next if route.threadpool.busy_size >= route.threadpool.pool_size
+
             delivery_info, properties, encoded_payload = queue.pop(route.queue_subscription_options)
             next unless encoded_payload # empty queue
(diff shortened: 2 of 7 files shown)


lib/action_subscriber.rb
```diff
@@ -81,6 +81,7 @@ def self.print_subscriptions
         logger.info "    --    exchange: #{route.exchange}"
         logger.info "    --       queue: #{route.queue}"
         logger.info "    -- routing_key: #{route.routing_key}"
+        logger.info "    --  threadpool: #{route.threadpool.pool_size}"
```
film42: I would like to make a cut at adding the threadpool name here, but I can't do it without some hacky code.

liveh2o: > ...Make the change easy..., then make the easy change – [Kent Beck](https://twitter.com/kentbeck/status/250733358307500032)

lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -17,6 +17,9 @@ def auto_pop!
         times_to_pop = [::ActionSubscriber::Threadpool.ready_size, ::ActionSubscriber.config.times_to_pop].min
         times_to_pop.times do
           queues.each do |route, queue|
+            # Handle busy checks on a per threadpool basis
+            next if route.threadpool.busy_size >= route.threadpool.pool_size
```
liveh2o: This is fine, but it could be easily rolled into the threadpool as a method called `busy?`:

```ruby
next if route.threadpool.busy?
```

liveh2o: Actually, it looks like you do have that method. Why not use it?

liveh2o: Scratch that. I was looking at the class method. But since there's a class method, that's all the more reason to add an instance method as well.

film42: This is currently a `Lifeguard::InfiniteThreadpool` so there's not much we can do. I plan to extend this in a separate PR and add `#busy?` as an instance method (as well as a threadpool name.. mentioned that above).

liveh2o:
:shipit:

abrandoned:
@film42 if you want to put a requirement on lifeguard > 0.0.8 then you can get the `busy?` and `name` methods for threadpools so you can do the things you want to do with the multi-pooling

film42:
Sweet! Now the startup message is a lot more clear. He's an example:
```
SomeSubscriber
  -- method: hit_low
    --    exchange: yolo
    --       queue: kevin.rogers.baseball.hit_low
    -- routing_key: rogers.baseball.hit_low
    --  threadpool: default, pool_size: 40
  -- method: hit
    --    exchange: yolo
    --       queue: kevin.rogers.baseball.hit
    -- routing_key: rogers.baseball.hit
    --  threadpool: some_other_pool, pool_size: 2
```

abrandoned:
:shipit:

film42:
Released as `2.1.2`.

PR #56, opened 2016-02-23
Configuration memo should return something truthy or this will change

film42:
cc @brianstien @mmmries

### DIFF
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -72,7 +72,7 @@ module ActionSubscriber
             ::ActionSubscriber.config.__send__("#{key}=", setting) if setting
           end
 
-          nil
+          true
         end
       end
     end


lib/action_subscriber/configuration.rb
```diff
@@ -72,7 +72,7 @@ def self.configure_from_yaml_and_cli(cli_options = {}, reload = false)
             ::ActionSubscriber.config.__send__("#{key}=", setting) if setting
           end
 
-          nil
+          true
```
film42: I could also return `self` here.

mmmries:
I have no preference around what gets returned. I don't imagine people will depend on the return value of this method. Maybe we should just stick with a literal `true` to make it as simple as possible

film42:
There's a bad spec that needs to be fixed. I already have the seed saved and will fix later. Re-ran travis and specs are passing.

mmmries:
:shipit:

PR #59, opened 2016-03-01
Change initialize to use not use Configuration

brettallred:
Configuration class doesn't have a configure method on it.  That method is on the ActionSubscriber Module


mmmries:
:+1: thanks for the catch

PR #62, opened 2016-04-08
Add option to have durable queues

brettallred:
Small patch to allow you to specify if a queue is durable.  It uses the Route API so you can simply set flag in your routes

```route FakeSubscriber, :foo, :durable => true```

The default is false to mimic the bunny and march hare gems and to be backwards compatible.

cc @mmmries @abrandoned @liveh2o @brianstien


film42:
:+1:

brettallred:
@film42 - do you guys not use durable queues and persistent messages?

mmmries:
@brettallred we don't use persistent messages or durable queues right now. We rely on the fact that the queues are redundant across a cluster to keep the schema of the queues and routing keys working.

I'm not sure of all the performance implications of using durable queues, but I don't see any reason not to support the flag. It seems like a no-brainer to me. One thing you might want to keep an eye out for in your production system is that if you try to declare a queue as durable, but the same already exists as transient I think you might get an error.

liveh2o:
:+1: for what @mmmries said.

mmmries:
This has been released as `2.4.0`. Thanks for the PR @brettallred

PR #68, opened 2016-09-09
track if we have acknowledged or not a message that needs it and if n…

abrandoned:
…ecessary reset the process when neither can be done

Make sure we are either rejecting or acknowledging outstanding jobs from rabbitmq, in the case where we can do neither we need to "restart" the process as rabbitmq will stop sending messages down to the process

we wanted to make this state very understandable to any user that experiences it so we added a verbose output message and an attempted graceful process shutdown

@film42 @quixoten @mmmries @brettallred

### DIFF
--- a/lib/action_subscriber/base.rb
+++ b/lib/action_subscriber/base.rb
@@ -61,17 +61,67 @@ module ActionSubscriber
     end
 
     def _at_least_once_filter
+      processed_acknowledgement = false
       yield
-      acknowledge
+      processed_acknowledgement = acknowledge
     rescue => error
       ::ActionSubscriber::MessageRetry.redeliver_message_with_backoff(env)
-      acknowledge
+      processed_acknowledgement = acknowledge
       raise error
+    ensure
+      rejected_message = false
+      rejected_message = reject unless processed_acknowledgement
+
+      if !processed_acknowledgement && !rejected_message
+        $stdout << <<-UNREJECTABLE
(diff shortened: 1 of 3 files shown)


mmmries:
Hmmm, now that Rails 5 has dropped and requires ruby > 2.2 should we just update all of our `.travis.yml` ruby verions? We could test on 2.2, 2.3, jruby-9.0.X and jruby-9.1.X?

lib/action_subscriber/base.rb
```diff
@@ -61,17 +61,73 @@ def acknowledge
...
+      rejected_message = reject unless processed_acknowledgement
+
+      if !processed_acknowledgement && !rejected_message
+        Process.kill(:TTIN, Process.pid)
+        Process.kill(:USR2, Process.pid)
+
+        $stdout << <<-UNREJECTABLE
+          CANNOT ACKNOWLEDGE OR REJECT THE MESSAGE
+
+          This is a exceptional state for RabbitMQ to enter and puts the current
```
mmmries: I would `RabbitMQ` to `ActionSubscriber`. We haven't gotten rabbitmq into a weird state, its just our own internal state that is now inconsistent.

mmmries:
:shipit: I think it would be best to put this in a pre-release version so we can roll it up in a few subscriber-heavy applications and let it sit in prod for a few days before it rolls out everywhere.

PR #69, opened 2016-09-12
synchronize channel actions that happen in the threadpool

mmmries:
Accessing channels from multiple threads for messsage acknowledgement will crash a channel and stop messages from being received on that channel this isn't very clear in the documentation for march_hare/bunny, but it is explicitly addressed here: https://github.com/rabbitmq/rabbitmq-java-client/issues/53

/cc @quixoten @abrandoned @brettallred @brianbroderick


lib/action_subscriber/synchronizer.rb
```diff
@@ -0,0 +1,15 @@
+require "thread"
+module ActionSubscriber
+  class Synchronizer
+    def initialize(delegate)
+      @delegate = delegate
+      @mutex = ::Thread::Mutex.new
```
abrandoned: maybe a ruby-2 thing, but thought Mutex was a top level class `::Mutex` ... ?

brianbroderick:
Looks good.  Thanks for looking into this.

lib/action_subscriber/march_hare/subscriber.rb
```diff
@@ -7,6 +7,12 @@ def cancel_consumers!
         march_hare_consumers.each(&:cancel)
       end
 
+      def create_queue(channel, queue_name, queue_options)
+        queue = ::MarchHare::Queue.new(channel, queue_name, queue_options)
+        queue.declare!
```
mmmries: those integration tests just saved my bacon again! It turns out that building a queue explicitly like this doesn't automatically declare it with the broker the way `channel.queue` does. 🌈

lib/action_subscriber/synchronizer.rb
mmmries: That must be a new thing. It looks like the top-level `Mutex` is documented in both [2.2.0](https://ruby-doc.org/core-2.2.0/Mutex.html) and in [1.9.3](https://ruby-doc.org/core-1.9.3/Mutex.html) so I'll use that instead.

abrandoned:
:shipit:

Issue #70, opened 2016-09-13
Shared Threadpools vs Multiple Consumers


mmmries:
The ActionSubscriber community has started to see issues with seeing subscribers (or in some cases just single queues/routes) freeze after operating for a while (see #66, #68 and #69).

After digging into this for a while it seem so to be related to the fact that message acknowledgement is not thread-safe. I found [a forum thread](https://groups.google.com/forum/#!topic/ruby-amqp/iOy1xLIkHvU) where Michael Klishin (the author of `march_hare`, `bunny` and a contributor to the core java client codebase) discussed this. I think the crux of his discussion for ActionSubscriber's use case boils down to:

> It says that you should not share channels between threads, specifically publishing. 
Having multiple consumers on a shared channel is fine, including if your consumers are 
concurrent, as long as you are careful with how you acknowledge (one-by-one should work fine, 
multiple-at-a-time will result in double-acks and channel termination).

We have some experiment underway right now where we could use a single thread to handle all message acknowledgement for an entire process of subscribers, a thread-per-channel for message acknowledgements. 

Considering all of these changes, I wondered what people's feelings would be about potentially removing the idea of shared threadpools entirely from ActionSubscriber?

An alternate design would be to allow a route to set a `number_of_threads => 5` option. This would register 5 consumers for that route and those 5 consumers could each run in their own thread. Locking a consumer to a thread requires the developer to allocate resources to each route at startup time, but it makes the mental model of the library a lot simpler as well.

I'm not sure this is a good direction for the library to take, but just wanted to get some feedback on the idea.

mmmries:
This was adopted for `3.0.0`. We now use multiple channels/consumers for concurrency in jRuby. In MRI we have a consumer work pool for each route with no shared threadpool.

PR #71, opened 2016-09-14
Action Subscriber 3.0 Proposal

mmmries:
Addresses #70 

The result of our recent explorations into thread safety issues has resulted in us finding out that the only safe place for us to send back message acknowledgements is within the consumer block that we register with `march_hare` (note: Bunny has slightly better concurrency guarantees, but we need to support both so we are using the lowest-common-denominator).

It turns out that the official rabbitmq drivers for C and Java both make this same recommendation of handling all messages form a single channel in a single thread. This means that for all `--mode=subscrbe` subscribers, shared threadpools managed by `action_subscriber` are off the table.

This PR is a proposal for a `3.0.0.pre1` version which will use `lifeguard` threadpools when in `pop` mode (polling), and will register multiple consumers (and support multiple connections) when in `subscribe` mode. What this means in essence is that instead drawing your routes like this:

```ruby
background_pool = ::ActionSubscriber::Threadpool.new_pool(:background, 10)
  default_routes_for ::JobSubscriber
  route ::JobSubscriber, :created
    :acknowledgements => true,
    :prefetch => 10,
    :queue => "me.other_service.jobs.created_low",
    :threadpool => background_pool
end
```

You will now specify this as:

```ruby
  default_routes_for ::JobSubscriber # this will now be a single threaded subscriber by default
  connection(:background, :threadpool_size => 10) do
    route ::JobSubscriber, :created,
      :acknowledgements => true,
      :concurrency => 10,
      :queue => "me.other_service.jobs.created_low",
  end
end
```

This allows you to control the concurrency level of each route, but also funnel the work into a shared threadpool (at the connection level).

> Note: for subscribers using the `pop` mode threadpools will still work just fine

Other changes proposed for 3.0:

* Remove all publisher code in favor of `active_publisher`
* Access to the connections is synchronized with a mutex (ie. ActionSubscriber::RabbitConnection.with_connect(:default){ |connection| do_some_work }`)
* Change `prefetch` to default value of 2 since each consumer will only do 1 job at at time now
* Remove the inferred routes fallback, clients must explicitly call `draw_routes` now
* Remove the attempted auto-loading of subscribers, we don't need this at all since we got rid of inferred routes

/cc @abrandoned @film42 @quixoten @brettallred @brianbroderick


mmmries:
BTW in order to confirm that new concurrency model I ran a test subscriber under both jRuby and MRI that had a connection (with 20 threads) and registered a route with `:concurrency => 20`. The route sleeps for 10 seconds at a time so I could easily check the logs to see that it started 20 jobs at a time, then after 10 seconds it finished all 20 jobs.  Rabbit shows 20 consumers registered in both cases and 20 unacknowledged messages until the whole queue was worked through.

brianbroderick:
@mmmries This all sounds great and we, at Nuvi, appreciate your efforts - we probably owe you lunch or something :).

Do you have any preliminary indications of how these proposed changes affect overall speed & efficiency?

mmmries:
@brianbroderick the tests I've done so far don't show any change in performance between the two concurrency models. The application that I was testing spends between 5 and 50ms on each job, and it's throughput after these changes was the same as it was before these changes: ~150 jobs/sec on a single queue. In this case my application is the bottleneck so I'm not sure how much the performance of the library itself has changed.

I'll see if I can do some smaller benchmarks where a simplistic application can handle a lot more throughput to see how the performance has changed.

abrandoned:
think this is a great step in the long-term success direction :shipit:

film42:
Great PR! :shipit:

mmmries:
I cut the pre-release gem from this branch because if I merge these changes into master, no one will be able to branch off of master for a while. `3.0.0.pre1` is released. Going to try to get it up into production in one of our apps today

mmmries:
looks like I have some cases where the process hangs on exist, I'll make sure we are cleaning up the connections so we don't get broken builds

lib/action_subscriber/rabbit_connection.rb
```diff
@@ -3,57 +3,40 @@
...
+      subscriber_connections.all?{|_name, connection| connection.connected?}
     end
 
-    def self.subscriber_connection
+    def self.subscriber_disconnect!
       SUBSCRIBER_CONNECTION_MUTEX.synchronize do
-        return @subscriber_connection if @subscriber_connection
-        @subscriber_connection = create_connection
+        subscriber_connections.each{|_name, connection| connection.close}
+        @subscriber_connections = []
```
abrandoned: seems odd that we would set this to `{}` on initialization and then `[]1 on disconnect; maybe we should include concurrent-ruby and use Concurrent::Map in both cases?

lib/action_subscriber/rabbit_connection.rb
```diff
@@ -3,57 +3,40 @@
...
-
-        @publisher_connection = nil
+        fail ArgumentError, "a #{name} connection already exists" if subscriber_connections[name]
+        subscriber_connections[name] = create_connection(settings)
       end
     end
 
     def self.subscriber_connected?
-      subscriber_connection.try(:connected?)
+      subscriber_connections.all?{|_name, connection| connection.connected?}
```
abrandoned: should we be using the mutex here as well?

lib/action_subscriber/middleware/env.rb
```diff
@@ -41,12 +41,14 @@ def initialize(subscriber, encoded_payload, properties)
...
 
       def acknowledge
+        fail ::RuntimeError, "you can't acknowledge messages under the polling API" unless @channel
         acknowledge_multiple_messages = false
         @channel.ack(@delivery_tag, acknowledge_multiple_messages)
         true
       end
 
       def reject
+        fail ::RuntimeError, "you can't acknowledge messages under the polling API" unless @channel
```
abrandoned: probably language should be "reject" here?

lib/action_subscriber/rabbit_connection.rb
mmmries: switched to use the mutex here

lib/action_subscriber/rabbit_connection.rb
mmmries: fixed

mmmries:
Just released `3.0.0.pre2`. We think it fixes a CPU utilization error under the `pop` mode with threadpools. Going to release it to a few of our internal services to test it under load.

mmmries:
Here are the list of remaining pain points that I'm aware of:

* Test suite hangs at on jRuby sometimes (all tests completed, but the process doesn't exit cleanly)
* The API between the binstub => `ActionSubscriber::Babou` => `ActionSubscriber` is pretty messy. I really want to clean this up and have a solid setup and tear-down procedure.

mmmries:
After 2 days of baking in production under load this is looking very stable. I'm going to go ahead and release the `3.0.0` version this morning. We can address the remaining painpoints in a `3.0.1`. I don't want to hold up people using the stable version of this library.

liveh2o:
:shipit:

PR #72, opened 2016-09-23
pass executor_factory to march_hare connections

mmmries:
The `threadpool_size` option isn't used by `march_hare` like I thought it was. Passing in an explicit `executor_factory` works. I tested on my local machine by setting up a threadpool of 100 threads and a route with concurrency of 100 and watched it on visualvm.

It is actually using 100 threads now. :+1:

I think this should be a quick `3.0.1` release

/cc @abrandoned @brianbroderick


abrandoned:
:shipit:

PR #73, opened 2016-09-25
Graceful shutdown

mmmries:
Fix graceful shutdown for the `3.X` series.

Now that we are not using the lifeguard threadpools for `--mode=subscribe`, we need to wait on the threadpools that are managed by bunny/march_hare. Rather than have a conditional about which mode we are in I decided to just wait on the lifeguard threadpools and then wait on the march_hare pool.

Interestingly `bunny` implements its own [threapool](https://github.com/ruby-amqp/bunny/blob/9ee43e6fb7c57aa40721551d6f43468dd99d871a/lib/bunny/consumer_work_pool.rb) and it has no interface for checking the state of the pool, or any thread-safe way for me to gracefully shut down their threadpools. So as a poor fallback I've implemented it to cancel the consumers (stop receiving new messages) and then wait for the full timeout to give as much time as possible for the threads to empty out.

> Note: I was about to start fixing the output for the initial printing of subscriptions with their threadpools and the output when you send a `USR2` signal. But there is a big problem where ActionSubscriber has no idea whether it is in `--mode=subscribe` or `--mode=pop`. This makes it really hard to print the right information, and rather than fix that I think the time has come to drop support for `--mode=pop`. We might want to drop that support first before merging this so we can simplify the implementation.

/cc @abrandoned @brianstien @brettallred @brianbroderick @quixoten

### DIFF
--- a/lib/action_subscriber/babou.rb
+++ b/lib/action_subscriber/babou.rb
@@ -73,17 +73,27 @@ module ActionSubscriber
       wait_loops = 0
       ::ActionSubscriber::Babou.stop_receving_messages!
 
-      while ::ActionSubscriber::Threadpool.busy? && wait_loops < ::ActionSubscriber.configuration.seconds_to_wait_for_graceful_shutdown
-        puts "waiting for threadpools to empty:"
-        ::ActionSubscriber::Threadpool.pools.each do |name, pool|
-          puts "  -- #{name} (remaining: #{pool.busy_size})" if pool.busy_size > 0
-        end
-        Thread.pass
+      puts "waiting for threadpools to empty (maximum wait of #{::ActionSubscriber.configuration.seconds_to_wait_for_graceful_shutdown}sec)"
+      loop do
         wait_loops = wait_loops + 1
+        any_threadpools_busy = false
+        ::ActionSubscriber::Threadpool.pools.each do |name, pool|
+          next if pool.busy_size <= 0
+          puts "  -- #{name} (remaining: #{pool.busy_size})"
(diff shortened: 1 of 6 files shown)


film42:
Seeing the following in travis for Jruby:

> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
> The build has been terminated

lib/action_subscriber/babou.rb
```diff
@@ -70,20 +70,12 @@ def self.stop_receving_messages!
...
-        Thread.pass
-        wait_loops = wait_loops + 1
-        sleep 1
-      end
-
-      puts "threadpool empty. Shutting down"
+      ::ActionSubscriber.wait_for_threadpools_to_finish_with_timeout(::ActionSubscriber.configuration.seconds_to_wait_for_graceful_shutdown)
+      puts "Shutting down"
+      ::Thread.new do
+        ::ActionSubscriber::RabbitConnection.subscriber_disconnect!
```
film42: Why does this need to be in its own thread?

mmmries: since we are in a `TRAP` context right here there are some things we can't really do (like acquire a mutex). I don't know why this limitation exists, but this is always how we have gotten around it.

film42: Oh! I didn't realize we were in a `TRAP` context.

film42:
Let's get this through before dropping pop mode so we leave people relying on pop mode a working implementation.

:shipit:

PR #74, opened 2016-09-26
make sure that at_most_once! and at_least_once! filters are applied f…

abrandoned:
…irst

Don't know if this is the behavior desired, but got burned by the`at_most_once!` filter being buried under some other filters that were trying to handle exceptions (and thus skipping the ack)

believe that at least the `at_most_once!` filter should be applied at the top and seems the same for `at_most_once!` as well; kept it as a filter, but forcibly apply it during execution instead of placing it wherever it may fall

@mmmries


mmmries:
Do the specs all pass locally for you on this branch? I'm seeing a few tests that are broken on all 4 travis builds.

I'm :+1: on pulling the enforcement of these flags into library code (separate of where the user's filters run), but if we are going to pull the enforcement of `at_least_once!` and `at_most_once!` I think we should skip calling the existing `_at_most_once_filter` filter and just replicate that behavior in the `run_action_with_filters` method.

mmmries:
Here's a rough pass at combining the logic into a single method

```ruby
def run_action_with_filters(env, action)
  processed_acknowledgement = false
  rejected_message = false

  processed_acknowledgement = acknowledge if @_at_most_once
  subscriber_instance = self.new(env)
  final_block = Proc.new { subscriber_instance.public_send(action) }

  first_proc = _around_filters.reverse.reduce(final_block) do |block, filter|
    Proc.new { subscriber_instance.send(filter, &block) }
  end
  first_proc.call
  processed_acknowledgement = acknowledge if @_at_least_once
rescue
   if @_at_least_once
     ::ActionSubscriber::MessageRetry.redeliver_message_with_backoff(env)
     processed_acknowledgement = acknowledge
   end
   raise
ensure
  rejected_message = reject if !processed_acknowledgement && @_acknowledge_messages
  if !rejected_message && !processed_acknowledgement && @_acknowledge_message
    $stdout << <<-UNREJECTABLE
      CANNOT ACKNOWLEDGE OR REJECT THE MESSAGE

      This is a exceptional state for ActionSubscriber to enter and puts the current
      Process in the position of "I can't get new work from RabbitMQ, but also
      can't acknowledge or reject the work that I currently have" ... While rare
      this state can happen.

      Instead of continuing to try to process the message ActionSubscriber is
      sending a Kill signal to the current running process to gracefully shutdown
      so that the RabbitMQ server will purge any outstanding acknowledgements. If
      you are running a process monitoring tool (like Upstart) the Subscriber
      process will be restarted and be able to take on new work.

       ** Running a process monitoring tool like Upstart is recommended for this reason **
    UNREJECTABLE

     Process.kill(:TERM, Process.pid)
  end
end
```

abrandoned:
@mmmries makes sense; I get what you are saying there, will refactor

mmmries:
One other idea we might consider here would be to move this functionality into a middleware piece. This idea has been considered twice in the past, but we haven't found a good implementation yet. Just a thought.

mmmries:

PR #75, opened 2016-09-26
Drop pop mode

mmmries:
This PR drops support for `--mode=pop` and removes all documentation and configuration that was related to `--mode=pop`. This is a backwards incompatible change and includes a note in the README about how to upgrade to this latest version.

I also contains updates to the startup messaging to give you better details about your routes. Here is an example from starting up under jRuby.

```
I, [2016-09-26T15:08:40.761000 #57064]  INFO -- : Loading configuration...
I, [2016-09-26T15:08:40.766000 #57064]  INFO -- : Requiring app...
I, [2016-09-26T15:08:41.081000 #57064]  INFO -- : Starting server...
I, [2016-09-26T15:08:41.256000 #57064]  INFO -- : Rabbit Hosts: ["localhost"]
Rabbit Port: 5672
Threadpool Size: 16
Low Priority Subscriber: false
Decoders:
  --application/json
  --text/plain

I, [2016-09-26T15:08:41.257000 #57064]  INFO -- : BaconSubscriber
I, [2016-09-26T15:08:41.258000 #57064]  INFO -- :   -- method: cooked
I, [2016-09-26T15:08:41.258000 #57064]  INFO -- :     --  connection: default (16 threads)
I, [2016-09-26T15:08:41.259000 #57064]  INFO -- :     -- concurrency: 16
I, [2016-09-26T15:08:41.259000 #57064]  INFO -- :     --    exchange: events
I, [2016-09-26T15:08:41.260000 #57064]  INFO -- :     --       queue: ron.bacon.cooked
I, [2016-09-26T15:08:41.260000 #57064]  INFO -- :     -- routing_key: bacon.cooked
I, [2016-09-26T15:08:41.261000 #57064]  INFO -- :     --    prefetch: 2 per consumer (32 total)
I, [2016-09-26T15:08:41.402000 #57064]  INFO -- : Action Subscriber connected
```

I will have a followup PR to this that will add:
* better stats reporting when we send a `USR2` signal to the process
* using bunny's consumer work pools rather than creating multiple consumer
* fixing the hanging jruby test suite

/cc @film42 @brianbroderick @abrandoned @quixoten


liveh2o:
Are we concerned about removing this functionality? Is it possible to override the prefetch number and enable acknowledgements under the hood to maintain "compatibility"?

mmmries:
@liveh2o I've checked all of our internal systems at MX and we aren't using this functionality anymore. I also checked with @brianbroderick at Nuvi and they aren't using this. I don't know of anyone using this mode, but its on rubygems so we can't be sure. I'm thinking we would remove this as a `4.0` release to signal that we had another backwards incompatible change.

The `prefetch` default is a pretty reasonable number for this behavior already. If we default to `acknowledgements => true` in the route we would probably also want to default to `at_most_once!` mode or something like that. I don't have any concrete reasons not to set a default there, I'm just not sure what the right acknowledgement default would be for new users.

liveh2o:
As a 4.0 release, I'm good with it. I just wasn't sure what the plan was.

action_subscriber.gemspec
```diff
@@ -26,7 +26,6 @@ Gem::Specification.new do |spec|
   else
     spec.add_dependency 'bunny', '>= 1.5.0'
   end
-  spec.add_dependency 'lifeguard', '>= 0.0.9'
```
film42: 👋

mmmries:
@zenovich I just remembered that you and I exchanged emails a while ago about using `action_subscribrer`. I just wanted to check with you if you are still using it or if you ever used the `--mode=pop`? We are planning to drop it in the next major release, but I would love to know if anyone was using it.

brianbroderick:
@mmmries @zenovich works with me.  I think we're good.

mmmries:

PR #76, opened 2016-09-26
Platform fixes

mmmries:
This is a follow on to #75

It adds much better information about the threadpools both during startup and when receiving a `USR2` signal.

~~I have one more commit I need to add to this which will drop our usage of `rabbitmq_http_api_client` so that we don't hang on travis runs anymore.~~

The hanging test runs turned out to be a threadpool leak in `march_hare`. It has [been fixed in that project](https://github.com/ruby-amqp/march_hare/pull/101). Once that gets released our builds will be fixed, but I went ahead and released a `4.0.0-pre1` version in the meantime.

/cc @abrandoned @film42 @liveh2o


mmmries:
I thought the fix for hanging test runs was simple, but it turns out to be related to https://github.com/ruby-amqp/march_hare/issues/97

I'll try to open a PR to resolve it there. In the meantime I think that this PR along with #75 put us in a good place to release a `4.0.0`. To check that assumption I would like to release a `4.0.0.pre1` version now and get it up into production to see how it does. Anyone want to belay that order?

belay that order

film42:
Go for the pre, yo.

mmmries:
Just as a heads up, we've had `4.0.0.pre1` in our production environment for several hours now and so far it is looking stable. We are using it in two different rails apps.  Going to let it go overnight to watch for slow memory leaks or other issues.

mmmries:

PR #78, opened 2016-11-17
allow username & password as cli parameters

zenovich:


mmmries:
have you tested that this parameters get passed through to the config? I don't remember off the top of my head exactly how we pass CLI parameters down to the config object that is used for the connection.

zenovich:
Yes, I'm passing username&password and an app becomes working. Without them, it doesn't.

mmmries:
Released as version `4.0.1`

PR #79, opened 2017-01-25
Add TLS support

jgaxn:
This PR allows bunny or march_hare to be configured for TLS.

@mmmries @abrandoned


mmmries:
This looks like a no-brainer to me. It sounds like there are still a few oddities and differences in behavior between how `bunny` and `march_hare` handle these options, but I think making them available is :+1: either way.

@film42 or @abrandoned anything you think we should verify or validate before releasing this?

mmmries:
released as [v4.1.0](https://rubygems.org/gems/action_subscriber/versions/4.1.0). I also added some notes to the README about the new configuration options.

PR #81, opened 2017-02-03
Add default values for TLS configuration options

jgaxn:
We don't pick up configuration options that don't have an entry in the `DEFAULTS` hash.

@mmmries @abrandoned @film42


mmmries:
:shipit:

looks good to me. It will be nice to have consistent defaults for these configuration values.

mmmries:
This was released as `v4.1.1`

PR #83, opened 2017-03-10
adding async connection mgmt and fixing query cache middleware

abrandoned:
fix issues with query cache middleware and add async connection mgmt ... make it the default in the middleware chain because it makes a signficant performance difference in reaping connections and the number of jobs that can get executed

@mmmries @film42


mmmries:
DANG! I was just about to open a PR that did this.

too fast

lib/action_subscriber/railtie.rb
```diff
@@ -3,10 +3,11 @@ class Railtie < ::Rails::Railtie
     config.action_subscriber = ::ActionSubscriber.config
 
     ::ActiveSupport.on_load(:active_record) do
+      require "action_subscriber/middleware/active_record/connection_management_async"
       require "action_subscriber/middleware/active_record/connection_management"
       require "action_subscriber/middleware/active_record/query_cache"
 
-      ::ActionSubscriber.config.middleware.use ::ActionSubscriber::Middleware::ActiveRecord::ConnectionManagement
+      ::ActionSubscriber.config.middleware.use ::ActionSubscriber::Middleware::ActiveRecord::ConnectionManagementAsync
```
mmmries: will changing this default cause problems for downstream dependencies? I'm thinking of despacho which will add its middleware before the `ConnectionManagement` middleware.

That should be pretty easy to fix, but it seems like this makes it a major version bump as opposed to a minor version bump.

abrandoned: @mmmries I am for a major version bump, are you?

mmmries: What if we just replaced the functionality in the existing middleware object (keep the same name). I don't know of any cases in which this will break for people using the library. It is only a performance improvement.

I'm okay with doing a major version bump or just replacing the existing class with the new implementation :+1:

mmmries:
Double-checked and the build failures are due to race conditions in the test suite. Ran it multiple times under MRI and jRuby locally to verify. This looks 👌

mmmries:
This was released as [4.2.1](https://rubygems.org/gems/action_subscriber/versions/4.2.1) and `4.2.1-java`

Issue #82, opened 2017-02-10
Warnings for drawing routes with inconsistent expectations of message acknowledgement


mmmries:
I've messed this up several times now where I made a subscriber with `at_most_once!` and forgot to pass `:acknolwedgements => true` in the route or vice versa.  It's always a pain to troubleshoot why the messages aren't going anywhere.

I think we should do a sanity check during the `draw_routes` phase to check for routes where the expectation of acknowledgements doesn't match what the subscriber expects. Doing it during `draw_routes` means that we'll usually find these while running unit tests even if action_subscriber isn't going to establish an actual connection.

If we find a mismatch we should make some noise.

![noise](https://1.bp.blogspot.com/-CE1n2fC3IXM/V3EwwmVaKaI/AAAAAAAApqo/TtaLbuz7PyoGFkFsAw7nLyUOeiG0vQcUACKgB/s400/Star%2BTrek%2BBeyond%2Btrailer%2B3%2BUSS%2BFranklin%2Bflaming%2Bswarm.gif)

mmmries:
This was resolved in #85

PR #85, opened 2017-04-26
Rbj/give warning on drawing routes

ryanbjones:
From https://github.com/mxenabled/action_subscriber/issues/82

This will show the warning when drawing the routes....since routes are not drawn during specs it would only be seen on booting up the server, but it should make debugging at that point easier

@mmmries @mattnichols @abrandoned @film42

### DIFF
--- a/lib/action_subscriber/bunny/subscriber.rb
+++ b/lib/action_subscriber/bunny/subscriber.rb
@@ -23,6 +23,9 @@ module ActionSubscriber
             logger.info "    -- routing_key: #{route.routing_key}"
             logger.info "    --    prefetch: #{route.prefetch}"
             logger.error "WARNING having a prefetch lower than your concurrency will prevent your subscriber from fully utilizing its threadpool" if route.prefetch < route.concurrency
+            if route.acknowledgements != subscriber.acknowledge_messages?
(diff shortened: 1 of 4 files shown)


mmmries:
Pulled this branch down and intentionally misconfigured a route and got the error message:

`I, [2017-04-26T11:21:22.525196 #60768]  INFO -- : Discrepancy in message acknowledgments in TransactionSubscriber`

:+1:, but the log level of `INFO` seems a bit misleading to me. Maybe this should be a `WARNING` or `ERROR` since we are pretty sure this will cause problems for the client code?

Also what do you think about doing this check right next to the warning we [have for probably incorrect `prefetch` settings](https://github.com/mxenabled/action_subscriber/blob/7d876a7a37d682df26775bed9a2f0251027ab0fe/lib/action_subscriber/bunny/subscriber.rb#L25)?

That warning shows right after printing the offending route rather than printing near the top of the log file. It might require doing it for both `march_hare` and `bunny`, but maybe it makes it easier to see that error?

ryanbjones:
@mmmries way better place to do it. Those changes are there now

mmmries:
Looks like `activerecord-5.1.0` dropped today and no longer has a `connection_id` method on `ActiveRecord::Base` which is causing the travis failures. This breakage is independent of these changes so I'm going go ahead an :shipit:

I'll try to get this released tomorrow morning and look into the `connection_id` issues

PR #87, opened 2017-07-20
add optional parameter for backoff schedule, fixes #86

### LINKED ISSUE #86: make message_retry schedule "configurable" in some way
https://github.com/mxenabled/action_subscriber/blob/master/lib/action_subscriber/message_retry.rb

Either  "global" config, or allow a single call to `redeliver_message_with_backoff` to take an optional argument that defines a custom retry schedule

some messages shouldn't jump from 12 seconds to 60 and then to 312 ... that's a long wait between retries

@mmmries

abrandoned:
include `retry` method (private) and make the backoff schedule an optional parameter in the retry method

@mmmries @film42

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,27 @@ Rabbit is told to expect message acknowledgements, but sending the acknowledgeme
 We send the acknowledgement right after calling your subscriber.
 If an error is raised your message will be retried on a sent back to rabbitmq and retried on an exponential backoff schedule.
 
+### retry
+
+A message can be sent to "retry" with `::ActionSubscriber::MessageRetry.redeliver_message_with_backoff` or the DSL method `retry` and optionally
+takes a "backoff schedule" which is a hash of backoff milliseconds for each retry, the default:
+
+```ruby
+  SCHEDULE = {
+    2  =>        100,
+    3  =>        500,
+    4  =>      2_500,
+    5  =>     12_500,
+    6  =>     62_500,
+    7  =>    312_500,
+    8  =>  1_562_500,
+    9  =>  7_812_500,
+    10 => 39_062_500,
+  }
+```
+
+when the schedule "returns" `nil` the message will not be retried
+
 Testing
 -----------------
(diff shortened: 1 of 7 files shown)


mmmries:
I like the implementation of passing a retry schedule as a hash. Makes it pretty simple to change the number of retries and the delay at the same time.

Do you think the the DSL `retry` method should reschedule and `acknowledge` the current message? Or keep those separate? I guess it could be weird if you have `at_least_once!` mode turned on and it tried to double acknowledge?

abrandoned:
@mmmries if we were to have the `retry` method `ack` then we would need to check which mode it is in and do the "most obvious" thing for that mode .... I didn't want the double ack scenario

your thoughts?

mmmries:
Since all of our `ack`/`reject` calls go through the `ActionSubscriber::Middleware::Env` object, we could record if an ack/reject already happened and prevent the double-ack scenario. But that feels a bit strange to me. It could mask a bug where someone is acknowledging a message earlier than they think.

Detecting the current mode seems overly complicated so I'm :+1: for just leaving the ack/reject to the user. If you are running in `at_most_once!` mode your message has already been acked so calling `retry` works fine, and if you are running in `at_least_once!` mode your message will get acked at the end of the method which is fine as well.

The only problem would be `at_least_once!` mode where you raise an error after calling `retry` since that would duplicate your message.

Let's just make sure we document that calling `retry` does not do anything with the ack/reject

mmmries:
Let's cut a pre-release from this branch and try it out in some part of the MX platform (sounds like grunt is where we want to use it first?)

Sorry if I'm being overly cautious, feel free to override if people feel differently. I just want to see it in use to verify that our assumptions work out the way we are thinking before I release it as a supported feature of the library.

mmmries:
:+1: for `redeliver` to avoid confusion with `begin..rescue..retry`

mmmries:
This has been tested multiple ways and run for multiple days in production with no issue so found. Going to merge and release as `4.3.0`

PR #88, opened 2017-07-24
Fix Shutdown for Jruby

film42:
Probably not a good idea to rely on the `at_exit` signal to rely on shutdown events. Jruby can handle kill signals, so I'm not sure why this behavior needed to be different. It was written this way 2 years ago, and I don't think whatever the reason was is actually relevant anymore.

I actually ran into this because we were seeing protobuf nats errors because it relies on an `at_exit` shutdown signal to cleanup the client connection. Since this hijacks the `at_exit`, it was messing things up bigly.

cc @abrandoned @mmmries

### DIFF
--- a/bin/action_subscriber
+++ b/bin/action_subscriber
@@ -40,11 +40,11 @@ module ActionSubscriber
       ::ActionSubscriber::Babou.start_subscribers
     end
   end
+end
 
-  [:INT, :QUIT, :TERM].each do |signal|
-    trap(signal) do
-      ::ActionSubscriber::Babou.stop_server!
-    end
+[:INT, :QUIT, :TERM].each do |signal|
+  trap(signal) do
+    ::ActionSubscriber::Babou.stop_server!
   end
 end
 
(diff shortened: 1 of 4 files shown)


abrandoned:
:shipit:

film42:
Going to cut a pre to make sure I'm not missing anything.

film42:
Looks like this will still trigger an `at_exit` too early.

film42:
@abrandoned I made a change to the shutdown sequence. It now behaves more like protobuf-nats https://github.com/abrandoned/protobuf-nats/blob/c6419365eab15f2815089a08100c1e44d4f35de5/lib/protobuf/nats/server.rb#L166-L176

mmmries:
👍  Let's cut a pre-release and test again. I think we had some of this done in a separate thread previously because we supported both a polling and a push mode, but we got rid of the polling mode so letting the shutdown happen in the main thread looks like a good idea.

film42:
Released pre `4.4.0.pre2`

film42:
This worked great in production. I tested shutdown with long running jobs. It properly unsubscribed, waited for work to complete before exiting the process/ main thread.

mmmries:
Let's release it as a stable version 💪

PR #89, opened 2017-08-09
Remove extra thread in error handler - already in a threadpool + dont allow error handler to re-raise errors

film42:
We're already off the main loop, so it's not protecting much. If it's trying to add extra inspection, it's probably not worth creating an extra thread for each message. 🔥 .

This also avoids letting the error handler re-raise errors, which seem to be the main cause for the java errors.

cc @mmmries @quixoten @liveh2o @brianstien @abrandoned

### DIFF
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -85,7 +85,13 @@ module ActionSubscriber
         'text/plain' => lambda { |payload| payload.dup }
       }
 
-      self.error_handler = lambda { |error, env_hash| raise }
+      self.error_handler = lambda do |error, env_hash|
+        logger = ::ActionSubscriber::Logging.logger
+
+        logger.error(error.message)
+        logger.error(error.class.to_s)
+        logger.error(error.backtrace.join("\n")) if error.try(:backtrace) && error.backtrace.is_a?(::Array)
+      end
 
       DEFAULTS.each_pair do |key, value|
         self.__send__("#{key}=", value)
--- a/lib/action_subscriber/middleware/error_handler.rb
+++ b/lib/action_subscriber/middleware/error_handler.rb
@@ -8,26 +8,10 @@ module ActionSubscriber
       end
 
       def call(env)
-        job_mutex = ::Mutex.new
-        job_complete = ::ConditionVariable.new
-
-        job_mutex.synchronize do
-          ::Thread.new do
-            job_mutex.synchronize do
-              begin
-                @app.call(env)
-              rescue => error
-                logger.error "FAILED #{env.message_id}"
-                ::ActionSubscriber.configuration.error_handler.call(error, env.to_h)
-              ensure
-                job_complete.signal
-              end
-            end
-          end
-
-          # TODO we might want to pass a timeout to this wait so we can handle jobs that get frozen
-          job_complete.wait(job_mutex)
-        end
+        @app.call(env)
+      rescue => error
+        logger.error "FAILED #{env.message_id}"
+        ::ActionSubscriber.configuration.error_handler.call(error, env.to_h)
       end
     end
   end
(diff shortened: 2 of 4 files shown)


film42:
Interesting. I see the error being raised in jruby. I'm not really sure how the thread wrapper is swallowing the error though.

film42:
Specs are passing. This is ready for review.

lib/action_subscriber/middleware/error_handler.rb
```diff
@@ -8,25 +8,18 @@ def initialize(app)
...
 
       def call(env)
-        job_mutex = ::Mutex.new
-        job_complete = ::ConditionVariable.new
+        @app.call(env)
+      rescue => error
+        logger.error "FAILED #{env.message_id}"
 
-        job_mutex.synchronize do
-          ::Thread.new do
```
mmmries: I think the reason we were wrapping these was because of the channel-killing behavior in march hare (also i think when we used to have lifeguard handle the threadpools, it would not restart crashed threads). So wrapping it in a thread avoids having an uncatchable error from killing your thread (ie java io ERRORs)

lib/action_subscriber/middleware/error_handler.rb
```diff
@@ -8,25 +8,18 @@ def initialize(app)
...
-                ::ActionSubscriber.configuration.error_handler.call(error, env.to_h)
-              ensure
-                job_complete.signal
-              end
-            end
-          end
-
-          # TODO we might want to pass a timeout to this wait so we can handle jobs that get frozen
-          job_complete.wait(job_mutex)
+        # There is more to this rescue than meets the eye. MarchHare's java library will rescue errors
```
mmmries: 10/10 will buy you at least 2 🌯 s the next time you are in town for leaving this comment

lib/action_subscriber/configuration.rb
```diff
@@ -85,7 +85,13 @@ def initialize
         'text/plain' => lambda { |payload| payload.dup }
       }
 
-      self.error_handler = lambda { |error, env_hash| raise }
+      self.error_handler = lambda do |error, env_hash|
+        logger = ::ActionSubscriber::Logging.logger
```
mmmries: will it cause a problem to memoize this value here? What if the user's code changes the logger after this file is loaded?

lib/action_subscriber/middleware/error_handler.rb
film42: MarchHare already has a built-in handler to convert java io errors: https://github.com/ruby-amqp/march_hare/blob/d568180074c24616f1240904d0a98d435d847a81/lib/march_hare/exceptions.rb#L99-L117

They'll still be caught here, though:
```
jruby-9.1.12.0 :013 >   error = java.io.IOException.new("yolo")
 => java.io.IOException: yolo                                                                                                                                                                                                                                                             
jruby-9.1.12.0 :014 > raise error rescue puts "we're good"
we're good                                                                                                                                                                                                                                                                                
 => nil    
```

lib/action_subscriber/configuration.rb
film42: 👍 Good thinking.

mmmries:
Now that we are using a more solid threadpool built into concurrent ruby and/or using a java threadpool under march hare, maybe we don't need to wrap things in a Thread for each item. That would definitely help performance. Maybe rather than trying to absolutely avoid the problem we should implement a [handle_cancel](https://github.com/ruby-amqp/march_hare/blob/d568180074c24616f1240904d0a98d435d847a81/lib/march_hare/consumers/base.rb#L30) callback in our march hare consumers, so that closing a channel and cancelling the consumers will trigger us to re-subscribe?

lib/action_subscriber/configuration.rb
film42: Oh wait, this is inside a lambda so the reference is never memoized. Sorry, spoke too soon.

film42:
Just cut an released `4.5.0.pre0`.

film42:
This has been running great in prod for nearly a week. I think it's ready to be released! 🌮

mmmries:
release it

film42:
Released!! 🎉 🌮

PR #91, opened 2017-10-19
Less channels

mmmries:
At MX our production cluster currently has 15k+ channels, 480 connections and 15k+ consumers. Each of these entities has an overhead associated with them. RabbitMQ tracks statistics around messages rates etc for each channel. We have some fairly strong evidence that reducing the number of channels and consumers will alleviate a lot of load for us.

This branch proposes that we rollback the DSL for configuring multiple connections and the DSL for setting a `:concurrency` value for each route. These are replaced by a `threadpool` DSL in the router. It allows the client application to create multiple threadpools and route messages to those threadpools. This has the side-effect of reducing numbers of connections since we are now managing our own threadpools instead of having `march_hare` manage threadpools. It also means that instead of having multiple channels and consumers to create concurrency we just have `prefetch`. So if you create a threadpool with 20 threads and put 2 routes in that threadpool that each have a prefetch of 20, action subscriber will receive jobs for both routes and feed them into the work queue for the threadpool where they will compete for space. This is very similar to the previous scheme of concurrency + connection threadpool settings, but we are using only a single channel per route.

If we put this change into just a single application in our production environment (grunt), we would reduce the total number of channels in the cluster from 15,358 to 8,674 channels.

This proposal also unifies the execution model so that the MRI and jRuby versions of the gem have the same concurrency setup and the same graceful shutdown process.

/cc @film42 @abrandoned @brianstien @liveh2o @quixoten


mmmries:
I also tested the thread-safety and performance of these changes by setting up a fake application that has a no-op subscriber with acknowledgements and a threadpool of 16 threads. This gave me the best chance of a race condition where two threads both try to acknowledge messages at the same time (simultaneous access to the channel). It processed hundreds of thousands of messages under both jRuby and MRI without issue. I also recorded throughput numbers and compared them to the master branch at various levels of prefetch values.

screen shot 2017-10-18 at 9 58 20 pm

There is a slight degradation in the max total throughput (more threads => more context switches), but our production use-case doesn't get anywhere close to 9k jobs/sec mark so this overhead shouldn't be a big deal for us.

lib/action_subscriber/rabbit_connection.rb
```diff
@@ -5,46 +5,27 @@ module RabbitConnection
...
+        @subscriber_connection = nil
       end
     end
 
-    def self.with_connection(name)
+    def self.with_connection
       SUBSCRIBER_CONNECTION_MUTEX.synchronize do
-        fail ArgumentError, "there is no connection named #{name}" unless subscriber_connections[name]
-        yield(subscriber_connections[name])
+        @subscriber_connection ||= create_connection
```
film42: Would we want to do something here like:
```
if @subscriber_connection &&  !@subscriber_connection.connected?
  @subscriber_connection.close # if this exists
  @subscriber_connection = nil
end

@subscriber_connection ||= create_connection
```
Or should we just wait for march_hare to reconnect?

mmmries: Right now we are relying on the march_hare/bunny reconnect logic so I think it makes sense to keep that behavior the same

mmmries:
I've added a commit which replaces the `connection` router method and the `:concurrency` setting. This makes it backward compatible with 4.X releases and provides a deprecation warning that explains the behavioral changes with links to an issue to provide feedback if people still need support for multiple connections. I've also cut a `5.0.0.pre0` release so we can move it up the chain at MX to test for stability.

jarecoo:
:shipit:

mmmries:
This has been in our production systems since last Friday and we haven't seen any indication of problems so far. One application switched `connection + concurrency` to `threadpool + prefetch` and the other didn't. Both worked fine.

I think this is ready to release as `5.0.0`. Any final thoughts before we release?

film42:
Nice work, @mmmries !!!

PR #93, opened 2017-10-30
Adds fix for projects using rspec-rails

asjaeger:


lib/action_subscriber/rspec.rb
```diff
@@ -1,5 +1,3 @@
-require 'rspec'
```
film42: Instead of removing this, let's require `rspec/core`. Better to be safe and explicit IMO. That's all the `rspec.rb` does anyway: https://github.com/rspec/rspec/blob/master/lib/rspec.rb

Thanks for reporting this, btw!

asjaeger: Awesome! Thank you for taking a look at this so quickly. I really appreciate it. I made the requested change.

film42:
Patch released in 5.0.1. I left a comment on #94 and closed it. ❤️

PR #95, opened 2017-12-07
make network recorvery interval configurable and default it to 1 second

abrandoned:
We want to make network recovery "longer" in our environment but keep the backward compat with the default being 1 second

(although if you use that you should be aware that it is a value that can have a race condition if it is too low)

https://github.com/ruby-amqp/march_hare/issues/83

This allows it to be configured by the configuration object/files

@film42 @quixoten

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,17 +125,26 @@ In an initializer, you can set the host and the port like this :
 Other configuration options include :
 
 * config.add_decoder - add a custom decoder for a custom content type
+* config.allow_low_priority_methods - Subscribe to `*_low` queues in addition to normal queues.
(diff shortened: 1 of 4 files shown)


jjcarstens:
Looks like `jruby-9.0.1.7` doesn't like the change. It [failed spec/integration/at_least_once_spec.rb:18](https://travis-ci.org/mxenabled/action_subscriber/jobs/313226733)

film42:
@jjcarstens There's a flappy spec we need to fix.

@abrandoned I updated the README to include the latest configuration options. Looks like we were missing quite a few things.

PR #96, opened 2017-12-08
add ack for multiple messages in at_most_once configuration

abrandoned:
We are finding that ack-ing every messages in `at_most_once!` mode is slowing down consumption; this is an example interface for allowing multiple messages to be ack'ed at once during `at_most_once!` mode .... certainly some questions around what happens when a batch of messages do not get ack'd together and how it can potentially starve the prefetch (also being sure that the prefetch is aligned correctly to this number so it doesn't starve ... ie a prefetch of 5 with an offset batch of 10 would deadlock)

@film42 @quixoten @brianstien

### DIFF
--- a/lib/action_subscriber/dsl.rb
+++ b/lib/action_subscriber/dsl.rb
@@ -9,10 +9,10 @@ module ActionSubscriber
       !!@_at_least_once
     end
 
-    def at_most_once!(multiple_offset = 1)
+    def at_most_once!(ack_every_n_messages = 1)
       @_acknowledge_messages = true
       @_at_most_once = true
-      @_multiple_offset = multiple_offset
+      @_ack_every_n_messages = ack_every_n_messages
     end
 
     def at_most_once?
@@ -55,8 +55,8 @@ module ActionSubscriber
       !!@_manual_acknowedgement
     end
 
-    def multiple_offset
-      @_multiple_offset
+    def ack_every_n_messages
+      @_ack_every_n_messages
     end
 
     def no_acknowledgement!
@@ -139,7 +139,7 @@ module ActionSubscriber
     def _run_action_at_most_once_multiple_with_filters(env, action)
       processed_acknowledgement = false
       rejected_message = false
-      if env.delivery_tag % multiple_offset == 0 # tags are monotonically increasing integers
+      if env.delivery_tag % ack_every_n_messages == 0 # tags are monotonically increasing integers
         processed_acknowledgement = env.acknowledge(true)
       else
         processed_acknowledgement = true # we are not acknowledging on this message and will wait for the offset to acknowledge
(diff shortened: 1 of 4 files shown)


lib/action_subscriber/dsl.rb
```diff
@@ -131,6 +136,41 @@ def _run_action_at_most_once_with_filters(env, action)
       end
     end
 
+    def _run_action_at_most_once_multiple_with_filters(env, action)
+      processed_acknowledgement = false
+      rejected_message = false
+      if env.delivery_tag % multiple_offset == 0 # tags are monotonically increasing integers
```
mmmries: one thing you might run into with keeping a copy of the delivery tag is that march hare will re-use a channel object when it recovers a broken connection, but your delivery tag is no longer valid for that channel.

In other words rabbitmq thinks that this is a new connection and a new channel, but your code is still holding the same channel object. MarchHare does this for convenience to the programmer, but if you try to acknowledge an old delivery tag form the previous version of the channel, it will raise an error.

This error can happen under the current version of `ActionSubscriber`, but batching your acknowledgements just widens the time-period in which an acknowledgement might fail and you have to consider the case that some of your `env`s came from a previous version of the channel.

abrandoned: @mmmries isn't that the same as today though? I only added an accessor to read the `delivery_tag` I didn't take a copy of it to store .... basically it will use the same deliver_tag it would have used, it will just use it less often and set the `multiple` header (if this is doing what I think it is)

mmmries: Oh... I see what you are doing here. This basically says to acknowledge all of the messages up to this current delivery tag? Digging through the java it looks like this is reflected at the protocol layer. So we are reducing the total back-and-forths between client and server.

This looks like it should work just fine for `at_most_once` mode.

mmmries:
To account for the possible deadlock issue, you could check for that when resolving routes. We have a [similar warning for cases where the route and the subscriber disagree about whether acknowledgements are expected](https://github.com/mxenabled/action_subscriber/blob/ac1636933a03e2fe0165dd3c1b3c42d7b761a03a/lib/action_subscriber/route_set.rb#L26-L28).

Separating the routing from the subscribing gives us a good way to reflect on this. So you could raise an error if you notice a route with a prefetch lower than the `multiple_offset` of the subscriber.

lib/action_subscriber/dsl.rb
```diff
@@ -9,9 +9,10 @@ def at_least_once?
       !!@_at_least_once
     end
 
-    def at_most_once!
+    def at_most_once!(multiple_offset = 1)
       @_acknowledge_messages = true
       @_at_most_once = true
+      @_multiple_offset = multiple_offset
```
mmmries: maybe `ack_every_n_messages` is a more descriptive name here?

lib/action_subscriber/middleware/env.rb
```diff
@@ -40,9 +41,9 @@ def initialize(subscriber, encoded_payload, properties)
         @subscriber = subscriber
       end
 
-      def acknowledge
+      def acknowledge(acknowledge_multiple_messages = nil)
```
film42: Seems like you can make this `acknowledge_multiple_messages = false` and remove the check below.

abrandoned:
well, this is not how multiple-ack works, if you end up acking with "multiple" encompassing the same set then the channel will blow up  ... (ie if you ack with multiple "23" and then ack with multiple "22" in another thread then it will blow the channel)

essentially would require in "greatest delivery tag" tracking per channel and then only acknowledging when it was "greater" than what has already been acked

(I implemented the above in a poorly constructed fashion, with a global, and it still doesn't change the performance of the subscriber)

PR #97, opened 2017-12-18
track duration of queue jobs flowing through

abrandoned:
want to get a handle on average timings for subscription types as they flow the the system, hopefully this is a start

@film42 @mmmries

### DIFF
--- a/lib/action_subscriber/middleware/router.rb
+++ b/lib/action_subscriber/middleware/router.rb
@@ -9,9 +9,9 @@ module ActionSubscriber
 
       def call(env)
         logger.info { "START #{env.message_id} #{env.subscriber}##{env.action}" }
-        started_at = ::Time.now
-        env.subscriber.run_action_with_filters(env, env.action)
-        ::ActiveSupport::Notifications.instrument("action_subscriber.duration.#{env.queue}", (::Time.now - started_at))
+        ::ActiveSupport::Notifications.instrument "process_event.action_subscriber", :subscriber => env.subscriber.to_s, :routing_key => env.routing_key, :queue => env.queue do
+          env.subscriber.run_action_with_filters(env, env.action)
+        end
         logger.info { "FINISHED #{env.message_id}" }
       end
     end
--- a/lib/action_subscriber/middleware/router.rb
+++ b/lib/action_subscriber/middleware/router.rb
@@ -8,9 +8,11 @@ module ActionSubscriber
       end
 
       def call(env)
-        logger.info "START #{env.message_id} #{env.subscriber}##{env.action}"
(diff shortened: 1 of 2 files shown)


mmmries:
@abrandoned I think this is already available via the `ActiveSupport` notifications we put in (see https://github.com/mxenabled/action_subscriber/blob/ac1636933a03e2fe0165dd3c1b3c42d7b761a03a/lib/action_subscriber/route_set.rb#L61-L65). These get picked up by `harness-action_subscriber` and reported to statsd. If you look in `grafana` for a chart called "ActionSubscriber" you should be able to narrow down the stats for any particular subscription across the platform.

abrandoned:
@mmmries I saw that, but thought that measured the time to push it on the threadpool (the average time for a db action was ~ 0.1ms which is crazy fast and a RPC only transaction is 0.06ms) ... that just seems way too fast (I can't even get a db connection and return from a db in < 200 usec)

mmmries:
I'm pretty sure that includes the time to execute the event. Maybe those subscribers have a shortcut so that they ignore messages in some cases and that drops their average way down?

You should be able to also get a `p95` of the latency from grafana and it should give you a better idea of how long most messages are taking.

abrandoned:
@mmmries it looks like it is just the time to push it on the threadpool
http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html#%3C%3C-instance_method

yeah, I pulled the p95 ... it only changes by about 20usecs (which is all the more strange)

mmmries:
https://github.com/mxenabled/harness-action_subscriber/blob/d5a19049d9cb6a0708580bfdefc88bc79dec7aff/lib/harness/action_subscriber.rb#L17-L21

the `processed_event` name in grafana was intended to be the latency for full executing the event

abrandoned:
below is the trans created that only calls sherlock, just way too fast for an rpc roundtrip on every call

image

mmmries:
Yeah, that can't possibly be the right median or 95th percentile time for a subscriber that makes and RPC roundtrip for every event. Maybe we broke this when we switched to using the `concurrent-ruby` threadpool instead of `lifeguard`?

abrandoned:
@mmmries yeah, I think so .... I think this resolves the mix-up

mmmries:
It would be nice to capture the latency of something heading into the pool as well, but as of this last commit this looks like a more correct form of measuring the "processing" time of each event.

It would probably be better to use a different name if you were measuring the time waiting + processing.

abrandoned:
@mmmries I agree, as I dig into instrumentation of the timing of "wait" to "execution" it is getting more important to know where things pause .... I'll put up some more instrumentation throughout the week

PR #99, opened 2018-01-13
Abrandoned/attempt to reject bad middlewares

abrandoned:
We have seen scenarios where the message is delivered but an error occurs before it is executed and the error handling in the subscriber never accounts for the message that is now "lost"

this adds a `nack` method (which is the recommended way of doing message rerun) and also `safe_` versions of all of the methods that need the channel and the delivery tag (the safe methods can be run multiple times and will only use the delivery_tag once)

This allows us to put in the error handler middleware a mechanism to make sure we `nack` a message that errors out

I also changed the `rescue` to rescue from `Exception` as that is the highest in the Ruby hierarchy so we can catch all errors/exceptions

Also makes sure we set the connection swapping and management middlewares with `insert_after` because they should be specifically included after the `Decoder` middleware

@film42 @mmmries @brianstien @quixoten @liveh2o

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ This also allows the broker to send messages as quickly as it wants down to your
 
 ### manual_acknowledgement!
 
-This mode leaves it up to the subscriber to handle acknowledging or rejecting messages. In your subscriber you can just call <code>acknowledge</code> or <code>reject</code>.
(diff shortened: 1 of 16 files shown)


film42:
Nice work! Code looks good. We’ll need to add an integration test highlighting the new nack feature but these are great finds and tweaks. Thanks for putting this together!

film42:
:shipit: Let's cut a pre and try this out!

PR #101, opened 2018-07-25
Fix YAML warning

sshock:
Fix so this warning won't show up anymore:
```
Called 'load' without the :safe option -- defaulting to safe mode.
You can avoid this warning in the future by setting the SafeYAML::OPTIONS[:default_mode] option (to :safe or :unsafe)
```

### DIFF
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -66,7 +66,12 @@ module ActionSubscriber
           yaml_config = {}
           absolute_config_path = ::File.expand_path(::File.join("config", "action_subscriber.yml"))
           if ::File.exists?(absolute_config_path)
-            yaml_config = ::YAML.load(::ERB.new(::File.read(absolute_config_path)).result, :safe => true)[env]
+            erb = ::ERB.new(::File.read(absolute_config_path)).result
+            if defined?(SafeYAML)
+              yaml_config = ::YAML.load(erb, :safe => true)[env]
+            else
+              yaml_config = ::YAML.load(erb)[env]
+            end
           end
 
           ::ActionSubscriber::Configuration::DEFAULTS.each_pair do |key, value|
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -66,7 +66,7 @@ module ActionSubscriber
           yaml_config = {}
           absolute_config_path = ::File.expand_path(::File.join("config", "action_subscriber.yml"))
           if ::File.exists?(absolute_config_path)
-            yaml_config = ::YAML.load(::ERB.new(::File.read(absolute_config_path)).result)[env]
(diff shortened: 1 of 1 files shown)


mmmries:
This looks good to me. @film42 do you want to double-check it since you still work at MX? Or maybe @liveh2o @brianstien or @abrandoned?

film42:
@sshock How does this flag make things more safe? Does it change how things are parsed? Just want to make sure this won't break anything. Change looks good, though.

sshock:
@film42 

My understanding is that it safe loading of YAML files avoids deserializing arbitrary objects, which keeps apps safe from certain exploits that are able to run arbitrary ruby code.

It should be noted that, with safe_yaml at least, which is what we always use, the default is already true, so we are already loading YAML files safely all the time.

So all this does it get rid of the warnings that we see all the time.

The only thing I don't know / understand is, can someone use action_subscriber with just regular yaml from ruby (not safe_yaml gem), and if so would this line turn into a syntax error?  I.e., would the regular YAML.load complain and fail, or would it just ignore this :safe option?

Since I'm new to ruby, that's the only part I'm not sure about, and if that is a problem how do we code this correctly to work with both regular yaml or safe_yaml?

liveh2o:
Unfortunately, this doesn't unless SafeYAML is loaded:

```irb
jruby-9.1.12.0 :004 > require 'yaml'
 => true
jruby-9.1.12.0 :005 > SafeYAML
NameError: uninitialized constant SafeYAML
	from org/jruby/RubyModule.java:3342:in `const_missing'
	from (irb):5:in `<eval>'
	from org/jruby/RubyKernel.java:1000:in `eval'
	from org/jruby/RubyKernel.java:1298:in `loop'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from /Users/adamhutchison/.rvm/rubies/jruby-9.1.12.0/bin/irb:13:in `<main>'
jruby-9.1.12.0 :006 > ::YAML.load("", :safe => true)
ArgumentError: unknown keyword: safe
	from /Users/adamhutchison/.rvm/gems/jruby-9.1.12.0@global/gems/psych-3.0.2-java/lib/psych.rb:6:in `load'
	from (irb):6:in `<eval>'
	from org/jruby/RubyKernel.java:1000:in `eval'
	from org/jruby/RubyKernel.java:1298:in `loop'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from /Users/adamhutchison/.rvm/rubies/jruby-9.1.12.0/bin/irb:13:in `<main>'
```

However, I think we may able to accommodate both like this:

```ruby
erb = ::ERB.new(::File.read(absolute_config_path)).result
if defined?(SafeYAML)
  yaml_config = ::YAML.load(erb, :safe => true)[env]
else
  yaml_config = ::YAML.load(erb)[env]
end
```

sshock:
sweet, I will update it accordingly

PR #102, opened 2018-10-31
we shouldnt need to create new channels on errors

ETetzlaff:
Currently redelivers are creating new channels for every call.  If we are getting ~5k errors a sec that's 1-1 for channel creation.  Instead, this reuses the existent channel to publish backoff messages.

### DIFF
--- a/lib/action_subscriber/message_retry.rb
+++ b/lib/action_subscriber/message_retry.rb
@@ -46,7 +46,7 @@ module ActionSubscriber
     end
 
     def self.with_exchange(env, ttl, retry_queue_name)
-      channel = RabbitConnection.with_connection{|connection| connection.create_channel}
+      channel = env.channel
       begin
         channel.confirm_select
         # an empty string is the default exchange [see bunny docs](http://rubybunny.info/articles/exchanges.html#default_exchange)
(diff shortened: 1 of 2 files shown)


film42:
Looks good to me!

@mmmries would you mind sanity checking this? Just want to make sure using the `env`'s channel isn't going to break anything. As far as we can tell it looks good, but just wanted a second opinion.

mmmries:
I'l try to take a look at it in the next day or two, slightly slammed at the moment, but the evenings should be a bit lighter

quixoten:
some of the [other code](https://github.com/mxenabled/action_subscriber/blob/85312611da8cd629c0c5e4a78137ea8d14dcab8f/lib/action_subscriber/middleware/env.rb#L62) suggests the channel won't always be defined. could that pose an issue?

mmmries:
@quixoten It looks like that code you referenced was for running `pop` mode (ie polling the broker rather than maintaining active connections) and we removed support for that in https://github.com/mxenabled/action_subscriber/pull/75

I can't think of any other case where `@channel` would be `nil` or somehow unusable for the purpose of publishing, but it's been a while since I thought about this library. I wouldn't consider my opinion definitive, but if its just a second (third) opinion then I would say 👍 . Cut a pre-release and put it into a production system somewhere for a little bit. Maybe try to force some errors. Then do a general release if all looks good.

ETetzlaff:
This is looking good in sand/qa.  Going to get this deployed into int and let it bake over night.

ETetzlaff:
This has been running in a prod app over the weekend and not seeing any issues.  @film42 What do you say to us merging this and that way we can move on with the other PR we've got queued up here?

film42:
:shipit: !!!

PR #103, opened 2018-11-06
rebuilds subscriptions on consumer_cancel_notify

hugh-j:
One of RabbitMQ's extensions to AMQP is to have a broker:
> send to the client a basic.cancel in the case of ... unexpected consumer cancellations.

[The documentation](https://www.rabbitmq.com/consumer-cancel.html) for the extension describes these "unexpected consumer cancellations" as:
> events, such as the queue being deleted, or in a clustered scenario, the node on which the queue is located failing, will cause the consumption to be cancelled, but the client channel will not be informed, which is frequently unhelpful.

The default behavior for both bunny and march_hare is to present support for this capability when connecting to RabbitMQ. Therefore, by default ActionSubscribers will receive basic.cancel signals when on the the scenarios described above occurs. Currently, when that happens, the ActionSubscriber will not attempt to resume consumption but rather will sit idle because no further messages will be sent from the broker on that channel. This change attempts to rebuild the subscription so consumption of messages will resume after the basic.cancel signal is received.

### DIFF
--- a/lib/action_subscriber/bunny/subscriber.rb
+++ b/lib/action_subscriber/bunny/subscriber.rb
@@ -7,6 +7,10 @@ module ActionSubscriber
         @bunny_consumers ||= []
       end
 
+      def set_bunny_consumers(consumers)
+        @bunny_consumers = consumers
+      end
+
       def cancel_consumers!
         bunny_consumers.each(&:cancel)
         ::ActionSubscriber::ThreadPools.threadpools.each do |name, threadpool|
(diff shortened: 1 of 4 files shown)


lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -51,6 +60,11 @@ def start_subscribers!
           end
           bunny_consumers << consumer
           queue.subscribe_with(consumer)
+      end
+
+      def start_subscribers!
```
liveh2o: This file was nice and alphabetized... Can we can bring that back?

lib/action_subscriber/bunny/subscriber.rb
```diff
@@ -24,14 +28,19 @@ def setup_subscriptions!
...
-        subscriptions.each do |subscription|
-          route = subscription[:route]
+      def start_subscription!(subscription)
+        route = subscription[:route]
           queue = subscription[:queue]
           channel = queue.channel
           threadpool = ::ActionSubscriber::ThreadPools.threadpools.fetch(route.threadpool_name)
           channel.prefetch(route.prefetch) if route.acknowledgements?
           consumer = ::Bunny::Consumer.new(channel, queue, channel.generate_consumer_tag, !route.acknowledgements?)
+          consumer.on_cancellation do |_basic_cancel|
```
liveh2o: This is awesome. W00t!

lib/action_subscriber/bunny/subscriber.rb
hugh-j: Done.

PR #104, opened 2018-11-28
More instrumentation around ack, nack, and reject operations

film42:
It occurred to me that we aren't measuring the time it takes to complete an ACK, NACK, or REJECT operation in action subscriber. We do this in active publisher and found it useful to detect if rabbitmq is pushing back on publishers to keep up. Does this happen on the subscriber side? I'm not sure, but this PR is me trying to find out.

I'm concerned about performance, but, since we're measuring a round-trip TCP message, I think the overhead will be negligible. Once I have the harness code in place, I can report back what kind of extra overhead we're talking about in practice.

cc @abrandoned @liveh2o @hugh-j @ETetzlaff

### DIFF
--- a/lib/action_subscriber/middleware/env.rb
+++ b/lib/action_subscriber/middleware/env.rb
@@ -3,6 +3,10 @@ require "securerandom"
 module ActionSubscriber
   module Middleware
     class Env
+      ACK_INSTRUMENT_KEY = "message_acked.action_subscriber".freeze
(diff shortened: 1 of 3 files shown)


abrandoned:
:shipit:

ETetzlaff:
LGTM

film42:
Cutting a pre-release. Rough benchmarks show the extra cost is almost negligible.

abrandoned:
@film42 are we ready to move on this? (and maybe some other MR's)

PR #105, opened 2018-12-13
set message id properly

ETetzlaff:
@film42 

I noticed we have a lot of weird spaces in the logs and came across this.  We aren't setting message_id appropriately as `properties.fetch(:message_id)` returns empty string and doesn't hit the SecureRandom.hex(3)

Alternatively, I think tying the delivery_tag of the message delivery and prepending that with the channel id/number might be a better solution for logging but this at least fixes a small symptom right now.

### DIFF
--- a/lib/action_subscriber/middleware/env.rb
+++ b/lib/action_subscriber/middleware/env.rb
@@ -38,7 +38,7 @@ module ActionSubscriber
         @has_been_nacked = false
         @has_been_rejected = false
         @headers = properties.fetch(:headers, {})
-        @message_id = properties.fetch(:message_id, ::SecureRandom.hex(3))
+        @message_id = properties[:message_id].presence || ::SecureRandom.hex(3)
         @queue = properties.fetch(:queue)
         @routing_key = properties.fetch(:routing_key)
         @subscriber = subscriber
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rvm:
  - 2.3.3
  - jruby-9.0.5.0
  - jruby-9.1.7.0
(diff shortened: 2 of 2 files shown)


lib/action_subscriber/middleware/env.rb
```diff
@@ -38,7 +38,7 @@ def initialize(subscriber, encoded_payload, properties)
         @has_been_nacked = false
         @has_been_rejected = false
         @headers = properties.fetch(:headers, {})
-        @message_id = properties.fetch(:message_id, ::SecureRandom.hex(3))
+        @message_id = properties.try(:message_id) ? properties.fetch(:message_id) : ::SecureRandom.hex(3)
```
liveh2o: You can use the new(ish) safe operator here and avoid the penalty of using `try`:

```ruby
@message_id = properties.&message_id ? properties.fetch(:message_id) : ::SecureRandom.hex(3)
```

However, I don't see how this will prevent us from capturing an empty string. Perhaps we should be [checking it's presence](https://api.rubyonrails.org/classes/Object.html#method-i-presence) instead?

```ruby
@message_id = properties.message_id.presence || ::SecureRandom.hex(3)
```

film42: :+1: for the `presence` check. I think the reason we're not receiving a message id is because we're not setting one on the publisher side.

ETetzlaff: @film42 That's a good point.  MessageID is an application use thing in AMQP.  I was thinking of how we might be able to make this useful.  We might want to keep this here, but I was thinking for logging purposes it might be more useful to have something like `channelID - deliveryTag`

PR #106, opened 2020-01-08
Add :if and :unless conditions to around_filter

newellista:
This will allow things like this:

  `around_filter :yell, :if => [:this_action]`

or

  `around_filter :yell, :unless => [:quiet_action]`

but, it won't handle:

  `around_filter :broken, :if => [:this_action, :that_action], :unless
  => [:this_action]`

That won't work.  

Just don't try it!  

It **will** fail silently!

### DIFF
--- a/lib/action_subscriber/dsl.rb
+++ b/lib/action_subscriber/dsl.rb
@@ -1,56 +1,54 @@
-require "pry"
-
-class Filter
-  attr_accessor :method
-  attr_accessor :included_actions
-  attr_accessor :excluded_actions
-
-  def initialize(method, options)
-    @method = method
-    @included_actions = @excluded_actions = []
-    parse_options(options)
-  end
-
-  def matches(action)
-    unless included_actions.empty?
-      return included_actions.include?(action)
-    end
-
-    unless excluded_actions.empty?
-      return false if excluded_actions.include?(action)
-    end
-
-    true
-  end
-
-private
-
-  def matches_excluded?(action)
-    return false if excluded_actions.empty?
-
-    return excluded_actions.include?(action)
-  end
-
-  def matches_included?(action)
-    return true if included_actions.empty?
-
-    return included_actions.include?(action)
-  end
-
-  def no_conditions
-    included_actions.empty? && excluded_actions.empty?
-  end
-
-  def parse_options(options)
-    return unless options
-
-    @included_actions  = options.fetch(:if, [])
-    @excluded_actions = options.fetch(:unless, [])
-  end
-end
 
 module ActionSubscriber
   module DSL
+    class Filter
+      attr_accessor :method
+      attr_accessor :included_actions
+      attr_accessor :excluded_actions
+
+      def initialize(method, options)
+        @method = method
+        @included_actions = @excluded_actions = []
+        parse_options(options)
+      end
+
+      def matches(action)
+        unless included_actions.empty?
+          return included_actions.include?(action)
+        end
+
+        unless excluded_actions.empty?
+          return false if excluded_actions.include?(action)
+        end
+
+        true
+      end
+
+    private
+
+      def matches_excluded?(action)
+        return false if excluded_actions.empty?
+
+        return excluded_actions.include?(action)
+      end
+
+      def matches_included?(action)
+        return true if included_actions.empty?
+
+        return included_actions.include?(action)
+      end
+
+      def no_conditions
+        included_actions.empty? && excluded_actions.empty?
+      end
+
+      def parse_options(options)
+        return unless options
+
+        @included_actions  = options.fetch(:if, [])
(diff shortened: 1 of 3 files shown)


lib/action_subscriber/dsl.rb
```diff
@@ -1,3 +1,54 @@
+require "pry"
```
film42: Let's remove the pry import.

lib/action_subscriber/dsl.rb
```diff
@@ -1,3 +1,54 @@
...
+
+class Filter
```
film42: This should probably live in the DSL namespace, right?

lib/action_subscriber/dsl.rb
newellista: D'oh!

Done

lib/action_subscriber/dsl.rb
```diff
@@ -95,7 +149,12 @@ def _run_action_with_filters(env, action)
       final_block = Proc.new { subscriber_instance.public_send(action) }
 
       first_proc = around_filters.reverse.reduce(final_block) do |block, filter|
-        Proc.new { subscriber_instance.send(filter, &block) }
+        if filter.matches(action)
+          p = Proc.new { subscriber_instance.send(filter.method, &block) } if filter.matches(action)
```
film42: I think you can drop the `p =` and `if filter.matches` since you already verified that in this if statement.

lib/action_subscriber/dsl.rb
```diff
@@ -1,5 +1,54 @@
...
+      attr_accessor :included_actions
+      attr_accessor :excluded_actions
+
+      def initialize(method, options)
+        @method = method
+        @included_actions = @excluded_actions = []
+        parse_options(options)
+      end
+
+      def matches(action)
```
film42: Is it worth memoizing the matching action here? Seems like a safe thing to do. Not sure if it's necessary, though.

lib/action_subscriber/dsl.rb
```diff
@@ -1,5 +1,54 @@
+
```
film42: Styling: Would you mind removing this line.

lib/action_subscriber/dsl.rb
```diff
@@ -1,5 +1,54 @@
...
+        included_actions.empty? && excluded_actions.empty?
+      end
+
+      def parse_options(options)
+        return unless options
+
+        @included_actions  = options.fetch(:if, [])
+        @excluded_actions = options.fetch(:unless, [])
+      end
+    end
```
film42: Would you mind adding a newline here?

lib/action_subscriber/dsl.rb
```diff
@@ -1,5 +1,54 @@
...
+        unless excluded_actions.empty?
+          return false if excluded_actions.include?(action)
+        end
+
+        true
+      end
+
+    private
+
+      def matches_excluded?(action)
```
film42: Looks like we can remove this and the next method.

lib/action_subscriber/dsl.rb
```diff
@@ -1,5 +1,54 @@
+
 module ActionSubscriber
   module DSL
+    class Filter
+      attr_accessor :method
```
film42: We should probably use a different name than method since method is a reserved name:
```
irb(main):001:0> class Yolo; end 
=> nil
irb(main):002:0> Yolo.new
=> #<Yolo:0x0000557f0a76da98>
irb(main):003:0> y = Yolo.new
=> #<Yolo:0x0000557f0a761680>
irb(main):004:0> y.method(:to_s)
=> #<Method: Yolo(Kernel)#to_s>
```

lib/action_subscriber/dsl.rb
newellista: Not quite sure what you mean.  `action` will change with each iteration, so we'd have to memoize for each action.  I don't see the value, unless I'm missing something?

film42: Yup, memoize on action. I don't think it's necessary now to optimize, because I don't think the gains will be massive, but just leaving a note.

lib/action_subscriber/dsl.rb
```diff
@@ -95,7 +133,11 @@ def _run_action_with_filters(env, action)
       final_block = Proc.new { subscriber_instance.public_send(action) }
 
       first_proc = around_filters.reverse.reduce(final_block) do |block, filter|
-        Proc.new { subscriber_instance.send(filter, &block) }
+        if filter.matches(action)
+          Proc.new { subscriber_instance.send(filter.method, &block) } if filter.matches(action)
```
film42: Should this and the around_filters_spec be `:action` as well? Maybe we should use something like `callback` here or `callback_method`. I think action vs method is really ambiguous. Sorry to keep nit-picking here.

PR #107, opened 2020-01-16
Accepting additional env properties in mock_subscriber

minond:
I might be working on a feature that requires using `env.action` within a subscriber and I noticed this value wasn't testable because it's hardcoded. This PR make that property along with the rest of the hardcoded environment properties configurable through the `mock_subscriber` method.


lib/action_subscriber/rspec.rb
```diff
@@ -62,12 +62,16 @@ def reject(delivery_tag, requeue_message)
...
+                                                        :channel,
                                                         :content_type,
                                                         :delivery_tag,
                                                         :exchange,
+                                                        :headers,
                                                         :message_id,
-                                                        :routing_key))
+                                                        :routing_key,
+                                                        :queue,
+                                                        :uses_acknoledgements))
```
film42: Should we add `uses_acknoledgements` to the shared_context below?

PR #108, opened 2020-05-15
Resubscribe on consumer cancellation

film42:
This adds a new feature flag (true by default) to automatically resubscribe when a cancellation event happens. This can happen when the queue is deleted, but it can also happen if a node in the cluster crashes without HA mode enabled. You can add the following to your config to disable these changes:

```yaml
resubscribe_on_consumer_cancellation: false
```

We should probably make a minor version bump with this release. It's different behavior than master but it's not a breaking change.

cc @abrandoned @minond


film42:
Released as 5.2.0

PR #113, opened 2022-01-05
Touch a file on start and rm the same file on stop

goutkannan:
Using the presence of subscriber-started file as an indicator for readiness.

### DIFF
--- a/lib/action_subscriber/babou.rb
+++ b/lib/action_subscriber/babou.rb
@@ -11,11 +11,11 @@ module ActionSubscriber
       ::ActionSubscriber.setup_subscriptions!
       ::ActionSubscriber.print_subscriptions
       ::ActionSubscriber.start_subscribers!
-      logger.info "Action Subscriber connected"
 
       ::FileUtils.mkdir_p(File.join(Rails.root, "tmp"))
       ::FileUtils.touch(File.join(Rails.root, "tmp", "subscriber-started"))
 
+      logger.info "Action Subscriber connected"
       while true
         sleep 1.0 #just hang around waiting for messages
         break if shutting_down?
@@ -26,6 +26,7 @@ module ActionSubscriber
       logger.info "Shutting down"
       ::ActionSubscriber::RabbitConnection.subscriber_disconnect!
       logger.info "Shutdown complete"
+      ::FileUtils.remove_file(File.join(Rails.root, "tmp", "subscriber-started"), force = true)
       exit(0)
     end
 
--- a/lib/action_subscriber/babou.rb
+++ b/lib/action_subscriber/babou.rb
@@ -1,4 +1,4 @@
-require "fileutils"
+require 'active_support/notifications'
 
 module ActionSubscriber
   module Babou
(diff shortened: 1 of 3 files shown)


lib/action_subscriber/babou.rb
```diff
@@ -9,8 +11,11 @@ def self.start_subscribers
       ::ActionSubscriber.setup_subscriptions!
       ::ActionSubscriber.print_subscriptions
       ::ActionSubscriber.start_subscribers!
-      logger.info "Action Subscriber connected"
 
+      ::FileUtils.mkdir_p(File.join(Rails.root, "tmp"))
+      ::FileUtils.touch(File.join(Rails.root, "tmp", "subscriber-started"))
```
film42, line 16: Seeing this I am realizing that we probably shouldn't make Rails a hard requirement here since it's not technically required to run action subscriber. How can we move this to the configuration.rb and railtie.rb to make this a generic feature? Maybe we should use active support instrumentation and add a callback in lib/action_subscriber.rb similar to how ruby-protobuf solves this problem?

film42: Could even be as simple as:
```
::ActiveSupport::Notifications.instrument("action_subscriber:server_started")

# lib/action_subscriber.rb
class ActionSubscriber
# ...
  def self.after_server_start(&block)
    ::ActiveSupport::Notifications.subscribe("action_subscriber:server_started") do |*args|
      block.call(*args)
    end
  end
# ...
end
```

goutkannan: Ok.  Pretty similar to proto.  First part of the comment was to move away from rails.root right ?

lib/action_subscriber/babou.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_support/notifications'
```
film42, line 1: Let's change this single quoted string to double quotes.

I think we need to rewrite CI to use circle-ci too. Bummer.

film42:
```
 (gk/touch_a_file_on_start) $ bx rake 
/home/film42/.rbenv/versions/2.7.4/bin/ruby -I/home/film42/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/rspec-core-3.10.1/lib:/home/film42/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/rspec-support-3.10.2/lib /home/film42/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/rspec-core-3.10.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb

Randomized with seed 3656
...........................................................................................................*...

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) ActionSubscriber::Middleware::Runner adds the router to the top of the stack
     # Not yet implemented
     # ./spec/lib/action_subscriber/middleware/runner_spec.rb:3


Finished in 11.75 seconds (files took 0.44433 seconds to load)
111 examples, 0 failures, 1 pending

Randomized with seed 3656

Coverage report generated for RSpec to /home/film42/Development/MX/action_subscriber/coverage. 719 / 876 LOC (82.08%) covered.
```

film42 (approved):

film42 (approved):

PR #117, opened 2022-04-27
Fix YAML loading

### LINKED ISSUE #116: Breaking changes in YAML loading (Psych)
[Psych](https://github.com/ruby/psych) (aka [YAML](https://github.com/ruby/yaml)) 4.x included a [breaking change](https://github.com/ruby/psych/issues/533) to how `YAML.load` works. In Psych 4.0, `load` calls `safe_load` under the hood, and is therefore "safe" by default, but that breaks configurations that support (among other things) aliases, which are disabled when using "safe" loading.

`YAML.unsafe_load` (and `YAML.unsafe_load_file`) is now the [canonical way](https://github.com/ruby/psych/issues/533#issuecomment-1019363688) to load trusted documents (i.e., config files).

To ensure maximum compatibility with old versions of Psych for ActionSubscriber, we also need to set a minimum version of Psych to ensure `unsafe_load` (and `unsafe_load_file`) is defined. The methods [were introduced](https://github.com/ruby/psych/commit/cb50aa8d3fb8be01897becff77b4922b12a0ab4c) in v3.3.2.

liveh2o:
Psych (aka YAML) 4.x included a breaking change to how `YAML.load` works In Psych 4.0, `load` calls `safe_load` under the hood, and is therefore "safe" by default, but that breaks configurations that support (among other things) aliases, which are disabled when using "safe" loading.

`unsafe_load` is now the canonical way to load trusted documents (i.e., config files): https://github.com/ruby/psych/issues/533#issuecomment-1019363688

To ensure maximum compatibility with old versions of Psych, we also need to set a minimum version of Psych to ensure `unsafe_load` is defined. The methods were introduced in v3.3.2: https://github.com/ruby/psych/commit/cb50aa8d3fb8be01897becff77b4922b12a0ab4c

Resolves #116

// @elbowdonkey @ETetzlaff @skunkworker

### DIFF
--- a/lib/action_subscriber/configuration.rb
+++ b/lib/action_subscriber/configuration.rb
@@ -69,11 +69,9 @@ module ActionSubscriber
           absolute_config_path = ::File.expand_path(::File.join("config", "action_subscriber.yml"))
           if ::File.exists?(absolute_config_path)
             erb = ::ERB.new(::File.read(absolute_config_path)).result
-            if defined?(SafeYAML)
-              yaml_config = ::YAML.load(erb, :safe => true)[env]
-            else
-              yaml_config = ::YAML.load(erb)[env]
-            end
+            # Defined in Psych 3.2+ and the new canonical way to load trusted documents:
+            # https://github.com/ruby/psych/issues/533#issuecomment-1019363688
+            yaml_config = ::YAML.unsafe_load(erb)[env]
           end
 
           ::ActionSubscriber::Configuration::DEFAULTS.each_pair do |key, value|
--- a/action_subscriber.gemspec
+++ b/action_subscriber.gemspec
@@ -1,7 +1,7 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
+lib = File.expand_path("../lib", __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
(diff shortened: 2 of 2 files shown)


lib/action_subscriber/configuration.rb
```diff
@@ -69,11 +69,9 @@ def self.configure_from_yaml_and_cli(cli_options = {}, reload = false)
...
           if ::File.exists?(absolute_config_path)
             erb = ::ERB.new(::File.read(absolute_config_path)).result
-            if defined?(SafeYAML)
-              yaml_config = ::YAML.load(erb, :safe => true)[env]
-            else
-              yaml_config = ::YAML.load(erb)[env]
-            end
+            # Defined in Psych 3.2+ and the new canonical way to load trusted documents:
+            # https://github.com/ruby/psych/issues/533#issuecomment-1019363688
+            yaml_config = ::YAML.unsafe_load(erb)[env]
```
liveh2o, line 74: Alternatively, we could check to see if `unsafe_load` is defined and not require a minimum version of Psych, but since Psych is a part of the standard library, I figured requiring a minimum of 3.3.2 would be fine.

skunkworker:
Looks good, lets do a prelease for testing.

skunkworker:
Looks good, lets do a prelease for testing.

PR #119, opened 2022-05-10
Remove minimum Psych constraint

### LINKED ISSUE #118: Minimum Psych version is incompatible with Rails < 6.1
Rails < 6.0 uses the old way of loading YAML files. This breaks loading database configs with aliases (a common development practice) with Psych 4.x. Adding a minimum version of Psych to the gemspec here forces an upgrade to Psych 4.x, which means we end up having to pin Psych to 3.3.2.

Rails 6.1 solves this problem by [checking](https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/secrets.rb#L29) if the new `unsafe_load` method exists, and falling back to `load` if it doesn't. That's a better way.

liveh2o:
Rails < 6.0 uses the old way of loading YAML files. This breaks loading database configs with aliases (a common development practice) with Psych 4.x. Adding a minimum version of Psych to the gemspec here forces an
upgrade to Psych 4.x, which means we end up having to pin Psych to 3.3.2.

Rails 6.1 solves this problem by checking if the new unsafe_load method exists, and falling back to load if it doesn't. That's a better way.

Resolves #118


lib/action_subscriber/configuration.rb
```diff
@@ -68,10 +68,10 @@ def self.configure_from_yaml_and_cli(cli_options = {}, reload = false)
           yaml_config = {}
           absolute_config_path = ::File.expand_path(::File.join("config", "action_subscriber.yml"))
           if ::File.exists?(absolute_config_path)
-            erb = ::ERB.new(::File.read(absolute_config_path)).result
+            erb_yaml = ::ERB.new(::File.read(absolute_config_path)).result
             # Defined in Psych 3.2+ and the new canonical way to load trusted documents:
             # https://github.com/ruby/psych/issues/533#issuecomment-1019363688
-            yaml_config = ::YAML.unsafe_load(erb)[env]
+            yaml_config = ::YAML.respond_to?(:unsafe_load) ? ::YAML.unsafe_load(erb_yaml)[env] : ::YAML.load(erb_yaml)[env]
```
skunkworker, line 74: is this safe for all psych versions? Should we have a more loose minimum constraint?

liveh2o: That is the most permissive as it uses the new method provided in 3.3.2+, and falls back to <3.1 behavior otherwise. Specs for this are passing under 2.6, 2.7, 9.2, 9.3, 3.0, and 3.1 (though other specs fail in 3+). The default in 2.6, and 2.7 is 3.1, and the default in 9.3 is 3.3.2. I think we're good.

minond (approved):

liveh2o:
> 

Yeah, it's a Faraday problem because of an old version. I have a branch that passes locally, but not in CI yet...

PR #120, opened 2022-06-01
Rails 6.1 local application name

kevinwatson:
The `parent_name` method has been replaced with `module_parent_name` in Rails 6.1. Let's add a check for the new method and also make it backwards compatible.

* https://apidock.com/rails/v6.0.0/Module/parent_name
* https://apidock.com/rails/v6.0.0/Module/module_parent_name

### DIFF
--- a/lib/action_subscriber/router.rb
+++ b/lib/action_subscriber/router.rb
@@ -58,7 +58,11 @@ module ActionSubscriber
                                  when ENV['APP_NAME'] then
                                    ENV['APP_NAME'].to_s.dup
                                  when defined?(::Rails) then
-                                   ::Rails.application.class.parent_name.dup
+                                   if ::Rails.application.class.respond_to?("module_parent_name")
+                                     ::Rails.application.class.module_parent_name.dup
+                                   else
+                                     ::Rails.application.class.parent_name.dup
+                                   end
                                  else
                                    raise "Define an application name (ENV['APP_NAME'])"
                                  end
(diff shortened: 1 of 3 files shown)


moveson (approved):
LGTM

moveson:
Just wondering if it's reasonable to add tests to exercise the code for this change. Pretty simple changes, so not a blocker, but might be nice to have some assurance or backward-compatibility.

lib/action_subscriber/router.rb
```diff
@@ -58,7 +58,11 @@ def local_application_name
                                  when ENV['APP_NAME'] then
                                    ENV['APP_NAME'].to_s.dup
                                  when defined?(::Rails) then
-                                   ::Rails.application.class.parent_name.dup
+                                   if ::Rails.application.class.respond_to?("module_parent_name")
```
film42, line 61: Nit: respond_to? should probably use a symbol arg.

kevinwatson: Good idea. Fixed.

Lancelittle:
This is needed for ruby3.0+/jruby9.4+ support. Worked for me locally.

film42:
Let’s make sure ruby 3 and jruby 9.4 are added to the test matrix.

kevinwatson:
> Let’s make sure ruby 3 and jruby 9.4 are added to the test matrix.

jruby 9.4 has been added. Ruby 3.0 and 3.1 are already in the mix: https://github.com/mxenabled/action_subscriber/blob/master/.circleci/config.yml#L59-L60

film42 (approved):

